PR #19, opened 2019-04-18
HotFix MT3

luca-fiorito-11:
Two problems were reported when sampling cross sections:
- command line option `--mt` was not working;
- when sampling MT=3, all daughter reactions were not perturbed accordingly.


luca-fiorito-11:
pull request on wrong branch. Will close!

PR #21, opened 2019-04-26
Develop docker

janmalec:
Hello! I have created a Dockerfile and attached a short text file with build instructions. Best, Jan


Dockerfile/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:18.04
+MAINTAINER Jan Malec (jan.malec@gmail.com)
+
+RUN apt update
+RUN apt install -y gfortran python3 python3-setuptools\
+    python3-numpy python3-dev unzip ca-certificates
+
+COPY sandy-develop.zip /
```
luca-fiorito-11: Instead of using `COPY` and a zip file, install `git` and run a clone command. :wink:

janmalec: Sorry for the late response, I will change the Dockerfile and push the update!

janmalec:
Hello! I have finally updated the Dockerfile as requested.

Issue #30, opened 2019-11-25
Integer initialization in rwfortran.f


Lindt8:
It seems that one of the "integer*4" declarations did actually need to be specifically set to length 4; 8 causes it to crash. (line 156 inside of "rilist")

luca-fiorito-11:
this should not be an issue anymore

PR #31, opened 2019-11-25
Fixed integer initialization issue in rwf

Lindt8:
It turns out that one of the integer variables (line 156) actually needs to be initialized as length 4.  Length 8 causes the same errors we had seen with some of the other variables where it had expected length 8 but received 4 except in reverse.

Resolves Issue #30 

(I'm not really sure what's going on with the differences with lines 17-29 and 69-75 being uncommented in what GitHub is saying is your current version of the file in the develop branch.  I pulled the file directly from your develop branch before editing it and can confirm both sets of lines are also commented in that file.)

### DIFF
--- a/fortran/rwfortran.f
+++ b/fortran/rwfortran.f
@@ -14,19 +14,19 @@ Cf2py integer intent(hide),depend(list) :: n=shape(list,0)
 c
 c
 c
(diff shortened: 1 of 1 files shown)


luca-fiorito-11:
outdated

PR #37, opened 2019-12-09
New class Endf6

luca-fiorito-11:
A new class `Endf6` in module `sandy.core.endf6` was created.

The new class is intended in the long term to replace the **outdated** class `sandy.formats.endf6.Endf6`.

Below is a list of major changes that were implemented:

- [x] a new module `sandy.core.records` was written to mimic the **outdated** module sandy.formats.records`. The major difference is that in the new module the dependence on fortran was removed.

- [ ] class `Errorr` was not yet rewritten

- [x] modules `sandy.formats.mf3` and  `sandy.formats.mf4` were rewritten under `sandy.sections.mf3` and `sandy.sections.mf4` to be compatible with the new class `Endf6`

- [ ] the same should be done for all other modules `mf1`, `mf5`, ...

- [x] unit tests were rewritten or adjusted

- [ ] a notebook `endf6_notebook` was started but must be finalized


luca-fiorito-11:
Removing dependence on fortran

PR #45, opened 2021-06-01
correct bug in fy.py that made notebooks fail 

luca-fiorito-11:


luca-fiorito-11:
- [x] correct bug in `fy.py`
- [ ] testing notebooks integrated in travis

PR #48, opened 2021-09-14
ENDF-B-VIII.0 library

AitorBengoechea:
updated the libraries.py and core/endf6.py rows so that they can obtain neutron nuclear data from the ENDF-B-VIII.0 library.


luca-fiorito-11:
This pull request is empty. Will close it!

luca-fiorito-11 (commented):
this is an empty file, probably a test

PR #49, opened 2021-09-14
Update to be able to read and write the mf9 section data

AitorBengoechea:


luca-fiorito-11:
@AitorBengoechea , there is a typo in `read_mf8`, it should be `read_mf9`.
Can you create examples for both functions `read_mf9` and `write_mf9` to make sure that these errors are detected?

luca-fiorito-11 (changes requested):
change the function `read_mf8` name into `read_mf9` and add examples

PR #89, opened 2021-11-18
Fy bmatrix perturbation and perturbation

AitorBengoechea:


sandy/fy.py
```diff
@@ -203,6 +210,92 @@ def _expand_zam(self):
...
+        Returns
+        -------
+        `Fy`
+            Fission yield instance with IFY calculated for a given value
+            of ZAM/E.
+
+        Examples
+        --------
+        >>> zam = [380900,551370,541350]
+        >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
```
luca-fiorito-11, line 276: you cannot test with a B-matrix for only 3 nuclides

sandy/core/endf6.py
```diff
@@ -87,8 +87,8 @@ def get_endf6_file(library, kind, zam, to_file=False):
             * 'nfpy' is a Neutron-Induced Fission Product Yields nuclear data
               file
             * 'decay' is a Radioactive Decay Data nuclear data file
-    zam : `int` or 'all'
-        zam = 'int' (individual nuclides)
+    zam : `int` or 'list' or 'all'
```
luca-fiorito-11, line 90: `list` or `'all'`

sandy/core/endf6.py
```diff
@@ -87,8 +87,8 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
+        zam = 'int' (individual nuclides) or 'list'(list of nuclides)
```
luca-fiorito-11, line 91: `list`

sandy/core/endf6.py
```diff
@@ -166,6 +166,10 @@ def get_endf6_file(library, kind, zam, to_file=False):
     >>> tape = sandy.get_endf6_file("endfb_71", 'nfpy', 'all')
     >>> assert type(tape) is sandy.Endf6
 
+    Import a list of Decay Data for JEFF-3.3.
+    >>> tape = sandy.get_endf6_file("jeff_33", 'decay', [380900,551370,541350])
```
luca-fiorito-11, line 170: PEP8

sandy/core/endf6.py
```diff
@@ -266,8 +270,16 @@ def get_endf6_file(library, kind, zam, to_file=False):
         text = "".join([foo_read(name, url) for name in files.values()])
         tape = Endf6.from_text(text)
     else:
-        filename = files[zam]
-        tape = foo_get(filename, url)
+        if type(zam) == list:
```
luca-fiorito-11, line 273: use `isinstance`.
I don't like this...what if instead of a list you get an array, or a tuple...it will break!

Check if `hasattr(zam, "__len__")` to see if it is an iterable

sandy/core/endf6.py
```diff
@@ -266,8 +270,16 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
+            filename = files[zam[0]]
+            tape = foo_get(filename, url)
+            for nuclide in zam[1:]:
+                filename = files[nuclide]
+                tape_2 = foo_get(filename, url)
+                tape = tape.merge(tape_2)
```
luca-fiorito-11, line 279: I prefer:
`tape = tape.merge(foo_get(filename, url))`

luca-fiorito-11 (changes requested):

sandy/fy.py
```diff
@@ -203,6 +210,92 @@ def _expand_zam(self):
...
+        `Fy`
+            Fission yield instance with given value ZAM/ZAP/MT/E perturbed.
+
+        Examples
+        --------
+        >>> tape = sandy.get_endf6_file("jeff_33",'nfpy','all')
+        >>> nfpy = Fy.from_endf6(tape)
+        >>> nfpy_pert = nfpy.custom_perturbation(922350, 459, 0.0253, 551370, -0.1)
+        >>> comp = nfpy_pert.data.query('ZAM==922350 & ZAP==551370 & MT==459 & E==0.0253').squeeze().FY
+        >>> assert np.setdiff1d(nfpy_pert.data.values,nfpy.data.values) == comp
```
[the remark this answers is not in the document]
AitorBengoechea: I have changed all the PEP 8 bugs, except for the too long line bugs. 
When I do numpy.setdiff1d(ar1, ar2, assume_unique=False), the outpur is: 1D array of values in ar1 that are not in ar2. So, the only value I have to return is the perturbed value. Test:
>>>a = np.array([[1,7,3],[9,5,6]])
>>>b = np.array([[1,2,3],[4,5,6]])
>>>np.setdiff1d(a, b)
array([7, 9])

sandy/fy.py
AitorBengoechea: I have changed the whole test, so that the IFYs of the 3 cores change to make it a valid test.

sandy/core/endf6.py
AitorBengoechea: I thought that with isinstance will not break because i introduce more ways to group data

AitorBengoechea:
@luca-fiorito-11 I update the version with the changes review that you did.

luca-fiorito-11:
@EnricaBelfiore , can you have a look?

sandy/core/endf6.py
```diff
@@ -15,6 +15,7 @@
 
 
 import pandas as pd
+import numpy
```
luca-fiorito-11, line 18: we usually import it as 
```
import numpy as np
```

sandy/core/endf6.py
```diff
@@ -87,8 +88,9 @@ def get_endf6_file(library, kind, zam, to_file=False):
             * 'nfpy' is a Neutron-Induced Fission Product Yields nuclear data
               file
             * 'decay' is a Radioactive Decay Data nuclear data file
-    zam : `int` or 'all'
-        zam = 'int' (individual nuclides)
+    zam : `int` or `list` or 'all'
+        zam = 'int' (individual nuclides) or `list` or `tuple` or 
+        `numpy.ndarray` (group of nuclides)
```
luca-fiorito-11, line 93: And what about a pandas.Series, or a Map object?
You should include all possible types of iterable...then the only way is to check if they have dunder `__len__`

sandy/core/endf6.py
```diff
@@ -266,8 +272,15 @@ def get_endf6_file(library, kind, zam, to_file=False):
         text = "".join([foo_read(name, url) for name in files.values()])
         tape = Endf6.from_text(text)
     else:
-        filename = files[zam]
-        tape = foo_get(filename, url)
+        if isinstance(zam, (list, tuple, numpy.ndarray)):
```
luca-fiorito-11, line 275: rewrite this according to the previous comment

sandy/core/endf6.py
```diff
@@ -266,8 +272,15 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
+            filename = files[zam[0]]
+            tape = foo_get(filename, url)
+            for nuclide in zam[1:]:
+                filename = files[nuclide]
+                tape = tape.merge(foo_get(filename, url))
+        elif type(zam) == int:
```
luca-fiorito-11, line 281: a simple else should be ok...what if ZAM is a float? e.g., 922350.0

sandy/fy.py
```diff
@@ -203,6 +213,140 @@ def _expand_zam(self):
...
+            Fission yield instance with IFY calculated for a given combination
+            of ZAM/e/decay_data.
+
+        Examples
+        --------
+        >>> zam = [591480, 591481, 601480]
+        >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
+        >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
+        >>> npfy = Fy(minimal_fytest_2)
+        >>> npfy_pert = npfy.apply_bmatrix(942390, 5.00000e+05, decay_fytest)
```
luca-fiorito-11, line 283: I still believe you cannot use the B-matrix this way....I mean, in this test

sandy/fy.py
```diff
@@ -203,6 +213,140 @@ def _expand_zam(self):
...
+            Energy to which calculations are to be applied.
+        decay_data : `sandy.DecayData`
+            Radioactive nuclide data for several isotopes.
+
+        Returns
+        -------
+        `Fy`
+            Fission yield instance with IFY calculated for a given combination
+            of ZAM/e/decay_data.
+
```
luca-fiorito-11, line 276: You are missing the note we said we had to add.

```
Notes
-----
.. note:: write your note
```

sandy/fy.py
```diff
@@ -203,6 +213,140 @@ def _expand_zam(self):
...
+        sensitivity = pd.DataFrame(C, index=B.index, columns=B.columns)
+        # Apply (1-B) matrix
+        fy_calc = sensitivity.dot(fy_data)
+        mask = (new_data.ZAM == zam) & (new_data.MT == 454) & (new_data.E == e)
+        new_data.loc[mask, 'FY'] = fy_calc.values
+        return self.__class__(new_data)
+
+    def apply_qmatrix(self, zam, energy, decay_data):
+        """
+        Perform CFY =(Q)(IFY) equation to calculate CFY in a given zam
```
luca-fiorito-11, line 307: CFY = Q * IFY

sandy/fy.py
```diff
@@ -203,6 +213,140 @@ def _expand_zam(self):
...
+
+        Returns
+        -------
+        `Fy`
+            Fission yield instance with IFY calculated for a given combination
+            of ZAM/e/decay_data.
+
+        Examples
+        --------
+        >>> zam = [591480, 591481, 601480]
```
luca-fiorito-11, line 328: same problem as for previous method

sandy/fy.py
```diff
@@ -203,6 +213,140 @@ def _expand_zam(self):
...
+
+        """
+        new_data = self.data.copy()
+        fy_data = self.filter_by('ZAM', zam)\
+            .filter_by('MT', 454)\
+            .filter_by("E", energy).data\
+            .set_index('ZAP')['FY']
+        Q = decay_data.get_qmatrix().loc[fy_data.index, fy_data.index]
+        # Apply qmatrix
+        fy_calc = Q.dot(fy_data).rename('FY').to_frame()
```
luca-fiorito-11, line 345: there is no need for renaming

luca-fiorito-11: Do you need to convert it to frame?

luca-fiorito-11 (changes requested):

sandy/fy.py
AitorBengoechea: The aim of the test is: 
Select related cores by decay (so that the B matrix is not zero) and create a variable (minimal_fytest_2) with these cores and invented IFY and CFY values. 
As IFY and CFY are invented, when applying bmatrix, all IFY values will change, since IFY and CFY are not related. In this way, by manually performing the multiplication between the matrix (1-B) and the vector CFY, we can check that the method results are correct.  
Finally, we compare the values of the original FY object and the one to which we have applied bmatrix. The only different values have to be the IFY values.

AitorBengoechea:
Luca do you see the changes I have made?
I'm working on my personal computer and I'm not sure if you can see the changes I've made. In my Github account if I log into the branch I am able to see them, but I'm not sure if in the pull request you can see them.

luca-fiorito-11:
@AitorBengoechea , yes I can see them!

EnricaBelfiore (approved):

sandy/core/endf6.py
```diff
@@ -266,8 +271,17 @@ def get_endf6_file(library, kind, zam, to_file=False):
         text = "".join([foo_read(name, url) for name in files.values()])
         tape = Endf6.from_text(text)
     else:
-        filename = files[zam]
-        tape = foo_get(filename, url)
+        if type(zam) == pd.Series:
+            zam = zam.values()
+        if zam.__len__() != 1:
```
luca-fiorito-11, line 276: @EnricaBelfiore, you shouldn't approve this! This will crash if zam is an integer.
@AitorBengoechea , I am surprised you don't get errors...probably because you don't have any test where you ask for a single decay nuclide (only all or list). This test should be added, otherwise these mistakes will not be detected.

sandy/core/endf6.py
```diff
@@ -266,8 +271,17 @@ def get_endf6_file(library, kind, zam, to_file=False):
         text = "".join([foo_read(name, url) for name in files.values()])
         tape = Endf6.from_text(text)
     else:
-        filename = files[zam]
-        tape = foo_get(filename, url)
+        if type(zam) == pd.Series:
```
luca-fiorito-11, line 274: I expected something like this example:

inputs
-------
```
url = sandy.libraries.URL_DECAY_JEFF_33_IAEA
files = sandy.libraries.DECAY_FILES_JEFF_33_IAEA
zam = (551340, 551330, 601480)
foo_get = sandy.Endf6.from_zipurl
```
what I would do
---------------
```
if hasattr(zam, "__len__"):
    tapes = map(lambda x: foo_get(files[x], url), zam)
    tape = reduce(lambda x, y: x.add_sections(y.data), tapes)
else:
    tape = foo_get(files[zam], url)
```
this should work both for scalar or iterables...lists, tuples, arrays, series, map objects, filter objects and so on.

It is taken from `Endf6.reduce`, as i mentioned to you!
Please check it before implementing it!

luca-fiorito-11 (changes requested):

sandy/core/endf6.py
AitorBengoechea: The tests for individual nucleus are put inside the get_endf6_file function, but even passing pytest with the mentioned error, pytest still pass the tests.
The program breaks in __FormattedFile.merge, i do not know why

sandy/core/endf6.py
```diff
@@ -271,17 +271,11 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
     else:
-        if type(zam) == pd.Series:
-            zam = zam.values()
-        if zam.__len__() != 1:
-            filename = files[zam[0]]
-            tape = foo_get(filename, url)
-            for nuclide in zam[1:]:
-                filename = files[nuclide]
-                tape = tape.merge(foo_get(filename, url))
+        if hasattr(zam, "__len__"):
```
luca-fiorito-11, line 274: Last comment, this section is better like this, right?!

Yours might fail in some specific case!

```
    if hasattr(zam, "__len__"):
        tapes = map(lambda x: foo_get(files[x], url), zam)
        tape = reduce(lambda x, y: x.add_sections(y.data), tapes)
    else:
        if str(zam).lower() == 'all':
            if kind.lower() == 'xs':
                raise ValueError("'all' option is not available for xs")
            text = "".join([foo_read(name, url) for name in files.values()])
            tape = Endf6.from_text(text)
        else:
            tape = foo_get(files[zam], url)
```

sandy/core/endf6.py
luca-fiorito-11: why I don't see `import numpy as np`?

Anyway, do not import numpy since you don't use it

sandy/core/endf6.py
luca-fiorito-11: can you check this?

luca-fiorito-11 (changes requested):

sandy/core/endf6.py
AitorBengoechea: In this order we have a problem: str variables has the atributte __len__

AitorBengoechea: So the program fail

luca-fiorito-11: ok, then keep it as it is!

luca-fiorito-11 (approved):

PR #90, opened 2021-11-23
Gls

AitorBengoechea:
@luca-fiorito-11  and @EnricaBelfiore 
I have uploaded WITHOUT ANY TEST the method to calculate the sensitivity for CFY and IFY according to GLS. I have also uploaded the custom perturbation method that introduces the perturbed data in the Fy object. 

If both of you think it is well structured (according to object-oriented programming) I will upload the tests. If you think there is a better structure, let me know and we will discuss it.


luca-fiorito-11:
Thanks Aitor! Will have a look.

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            Fission yield instance with IFY and CFY perturbed using GLS
+            for a given ZAM/e/decay_data.
+
+        """
+        new_data = self.data.copy()
+        mask = (new_data.ZAM == zam) & (new_data.E == e)
+        # Divide the CFY and IFY:
+        fy_type = perturbed_fy.data.loc[mask].groupby('MT')
+        IFY_pert, CFY_pert = [fy_type.get_group(x)[['ZAP', 'FY']]
+                              .set_index('ZAP') for x in fy_type.groups]
```
luca-fiorito-11, line 482: it's not really clear what you do here...I guess the syntax is too complex.
Try simplify it! Use maybe more steps, better variable names, comments,...

maybe this could help simplify, but I still don't understand what you are doing here...
```
groups = perturbed_fy.data.loc[mask].groupby('MT')
IFY_pert, CFY_pert = [group[['ZAP', 'FY']].set_index('ZAP') for mt, group in groups]
```

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        fy_type = new_data.loc[mask].groupby('MT')
+        IFY, CFY = [fy_type.get_group(x)[['ZAP', 'FY']]
+                    .set_index('ZAP') for x in fy_type.groups]
+        # perform the GLS method
+        if pert_data == 454:
```
luca-fiorito-11, line 487: Let's remove this part...we will never use GLS to update the cumulative yields

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            CFY_new = CFY + sensitivity.dot(IFY_pert-IFY)
+            CFY_new[CFY_new < 1.0e-14] = 0  # Avoid numerical fluctuations
+            CFY = CFY_new
+            IFY = IFY_pert
+        elif pert_data == 459:
+            IFY_new = IFY + sensitivity.dot(CFY_pert-CFY)
```
luca-fiorito-11, line 493: Now I understand better what you do...

I think a fundamental issue is that you use variable name `CFY_pert`, while this does not need to be a *perturbed* parameter

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            IFY_new[IFY_new < 1.0e-14] = 0
```
luca-fiorito-11, line 494: should this threshold be a keyword argument?

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            IFY = IFY_new
+            CFY = CFY_pert
```
luca-fiorito-11, line 496: why these equalities?

sandy/fy.py
```diff
@@ -357,6 +360,147 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            Perturbed data MT is not in Fy object.
+
+        Returns
+        -------
+        `Fy`
+            Fission yield instance with IFY and CFY perturbed using GLS
+            for a given ZAM/e/decay_data.
+
+        """
+        new_data = self.data.copy()
```
luca-fiorito-11, line 477: it is just better to call it `data`

luca-fiorito-11 (changes requested):
A few general comments:

- let's try and use `CategoryCov` rather than `_Cov`, which by the way will disappear
- a lot of lines where you operate on the covariance matrix could be converted into methods
- can we make `custom_GLS` even more general where updates could also come from something that is not CFYs? with different sensitivities...

sandy/core/cov.py
```diff
@@ -399,14 +406,39 @@ def glls(self, S, Vy):
...
-    def sandwich(self, Si, Sj=None):
+    def sandwich(self, S):
+        """
+        Apply the sandwich formula to the CategoryCov object for a given
+        pandas.Series.
+
+        Parameters
+        ----------
+        S : `pd.Series`
+            Standard desviations.
```
luca-fiorito-11, line 417: these are in general sensitivities, not standard deviations

sandy/core/cov.py
```diff
@@ -399,14 +406,39 @@ def glls(self, S, Vy):
...
+        >>> var = pd.Series([1, 2, 3])
+        >>> cov = sandy.CategoryCov(S)
+        >>> cov.sandwich(var)
+                    0           1           2
+        0 1.00000e+00 0.00000e+00 0.00000e+00
+        1 0.00000e+00 8.00000e+00 0.00000e+00
+        2 0.00000e+00 0.00000e+00 2.70000e+01
+        """
+        if self.data.index.values.all() != S.index.values.all():
+            raise sandy.Error("The index are not the same")
```
luca-fiorito-11, line 437: "the indices of the sensitivity array and of the `CategoryCov` are not the same"

sandy/core/cov.py
```diff
@@ -399,14 +406,39 @@ def glls(self, S, Vy):
...
+        >>> S = np.diag(np.array([1,2,3]))
+        >>> var = pd.Series([1, 2, 3])
+        >>> cov = sandy.CategoryCov(S)
+        >>> cov.sandwich(var)
+                    0           1           2
+        0 1.00000e+00 0.00000e+00 0.00000e+00
+        1 0.00000e+00 8.00000e+00 0.00000e+00
+        2 0.00000e+00 0.00000e+00 2.70000e+01
+        """
+        if self.data.index.values.all() != S.index.values.all():
```
luca-fiorito-11, line 436: why don't you try [`pandas.Index.equals`](https://pandas.pydata.org/docs/reference/api/pandas.Index.equals.html)?

sandy/core/cov.py
```diff
@@ -399,14 +406,39 @@ def glls(self, S, Vy):
...
         C = self.data.values
-        Si_ = Si.reshape(self.size, -1)
-        if Sj is None:
-            Sj_ = Si_
-        else:
-            Sj_ = Sj.reshape(self.size, -1)
-        return Si_.T.dot(C.dot(Sj_))
+        s = S.values
+        out = corr2cov(C, s)
+        return self.__class__(out)
```
luca-fiorito-11, line 441: watch out here...your output `CategoryCov` will not have anymore the index and columns of the original `CategoryCov`, but just a list of integers.

Can you add a test to check this?

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
     -------
     `numpy.ndarray`
         covariance matrix
+
+    Examples
+    --------
+    >>> S = np.array([1,2,3])
```
luca-fiorito-11, line 990: PEP8

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
...
+    --------
+    >>> S = np.array([1,2,3])
+    >>> var = np.array([[1, 0, 2, 0], [0, 3, 0, 0], [4, 0, 5, 0], [0, 0, 0, 0]])
+    >>> corr2cov(var,S)
+    array([[ 1,  0,  6,  0],
+           [ 0, 12,  0,  0],
+           [12,  0, 45,  0],
+           [ 0,  0,  0,  0]])
+
+    >>> S = np.array([1,2,3])
```
luca-fiorito-11, line 998: PEP8

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
...
+    >>> var = np.array([[1, 0, 2], [0, 3, 0], [4, 0, 5]])
+    >>> corr2cov(var,S)
```
luca-fiorito-11, line 1000: PEP8

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
     -------
     `numpy.ndarray`
         covariance matrix
+
+    Examples
+    --------
+    >>> S = np.array([1,2,3])
+    >>> var = np.array([[1, 0, 2, 0], [0, 3, 0, 0], [4, 0, 5, 0], [0, 0, 0, 0]])
+    >>> corr2cov(var,S)
```
luca-fiorito-11, line 992: PEP8

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
...
+    array([[ 1,  0,  6],
+           [ 0, 12,  0],
+           [12,  0, 45]])
     """
     dim = corr.shape[0]
-    S = np.repeat(s, dim).reshape(dim, dim)
-    return S.T * (corr * S)
+    if len(s) > dim:
+        raise sandy.Error("The shape of the variables is not correct")
```
luca-fiorito-11, line 1007: I would like to stop using `sandy.Error`...maybe use `TypeError`?
you can find [here](https://docs.python.org/3/library/exceptions.html#bltin-exceptions) a list of errors that you can use

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
...
+    if len(s) != dim:
+        dim_ext = abs(len(s)-dim)
```
luca-fiorito-11, line 1009: PEP8

sandy/core/cov.py
```diff
@@ -951,10 +984,33 @@ def corr2cov(corr, s):
...
+        S = np.pad(np.diag(s), ((0, dim_ext), (0, dim_ext)))
```
luca-fiorito-11, line 1010: can you add a comment about what you do here? it is not clear

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: When not entered in the index as such, this function breaks. By this I mean that with indexes like pd.Index([1, 2, 3]) it works, but in the example the index is not specified [RangeIndex(start=0, stop=3, step=1)], and in this case it does not work. 

However, I'll add a test to check the case pd.Index([1, 2, 3])

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+
+        Parameters
+        ----------
+        nonzero_idxs : numpy.ndarray
+            The indices of the diagonal that are not null.
+        cov_reduced : sandy.core.cov._Cov
+            The reduced matrix.
+        dim : int
+            Dimension of the original matrix.
+
```
luca-fiorito-11, line 394: please, add a note

```
Notes
-----
.. note:: this method was developed to be used after calling `_reduce_size`
```

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+        """
+        Restore the size of the matrix
+
+        Parameters
+        ----------
+        nonzero_idxs : numpy.ndarray
+            The indices of the diagonal that are not null.
+        cov_reduced : sandy.core.cov._Cov
+            The reduced matrix.
+        dim : int
```
luca-fiorito-11, line 392: `int`

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+            Dimension of the original matrix.
+
+        Returns
+        -------
+        cov : sandy.core.cov._Cov
```
luca-fiorito-11, line 397: no, this is a `np.ndarray`, I guess

sandy/core/cov.py
luca-fiorito-11: apply for all other parameters

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+
+    def _restore_size(cls, nonzero_idxs, cov_reduced, dim):
+        """
+        Restore the size of the matrix
+
+        Parameters
+        ----------
+        nonzero_idxs : numpy.ndarray
+            The indices of the diagonal that are not null.
+        cov_reduced : sandy.core.cov._Cov
```
luca-fiorito-11, line 390: this is a pd.Dataframe , right?

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+            The reduced matrix.
+        dim : int
+            Dimension of the original matrix.
+
+        Returns
+        -------
+        cov : sandy.core.cov._Cov
+            Matrix of specified dimensions.
+
+        """
```
luca-fiorito-11, line 400: can you add an example?

sandy/core/cov.py
luca-fiorito-11: but it can be something else...maybe a `CategoryCov `. In that case it should be a classmethod

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+        >>> reduce_matrix
+                      1	          2
+        1	2.00000e+00	0.00000e+00
+        2	0.00000e+00	3.00000e+00
+        """
+        nonzero_idxs = np.flatnonzero(np.diag(self.data))
+        cov_reduced = self.data.loc[nonzero_idxs, nonzero_idxs]
+        return nonzero_idxs, cov_reduced
+
+    def _restore_size(cls, nonzero_idxs, cov_reduced, dim):
```
luca-fiorito-11, line 382: why `cls` and not `self`?

luca-fiorito-11: you don't use `self`, then you can remove it and make it a static method.
But, as reported below, maybe it should be a class method

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+        cov : sandy.core.cov._Cov
+            Matrix of specified dimensions.
+
+        """
+        cov = np.zeros((dim, dim))
+        for i, ni in enumerate(nonzero_idxs):
+            cov[ni, nonzero_idxs] = cov_reduced[i]
+        return cov
+
+    def inv(self):
```
luca-fiorito-11, line 406: I'd rather call it `invert`

sandy/core/cov.py
```diff
@@ -341,6 +343,100 @@ def corr(self):
...
+                    0           1           2
+        0 0.00000e+00 0.00000e+00 0.00000e+00
+        1 0.00000e+00 5.00000e-01 0.00000e+00
+        2 0.00000e+00 0.00000e+00 3.33333e-01
+        """
+        index, columns = self.data.index, self.data.columns
+        M_nonzero_idxs, M_reduce = self._reduce_size()
+        unit = np.identity(len(M_reduce))
+        M_reduce_inv = splu(csc_matrix(M_reduce)).solve(unit)
+        data = self._restore_size(M_nonzero_idxs, M_reduce_inv, len(self.data))
```
luca-fiorito-11, line 435: make sure to correct if you change restore size to a classmethod...and also if you change the output type

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
         Q, R = scipy.linalg.qr(M.T)
         return R.T
 
-    def glls(self, S, Vy):
+    @classmethod
+    def get_covariance(cls, std):
```
luca-fiorito-11, line 500: this name is unclear...it would be ok if this method was assigned to the FY object.
But it is a CategoryCov method (which is fine).
Then `get_covariance` doesn't mean anything.

Maybe `from_stdev`? or `from_var`?

or maybe you can create both methods

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
 
-    def glls(self, S, Vy):
+    @classmethod
+    def get_covariance(cls, std):
+        """
+        Construct the covariance matrix from the variance vector.
+
+        Parameters
+        ----------
+        std : `pandas.Series`
```
luca-fiorito-11, line 506: don't call it `std` if it is a variance

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+        >>> cov = sandy.CategoryCov.get_covariance(S)
+        >>> cov
+                    1           2           3
+        1 0.00000e+00 0.00000e+00 0.00000e+00
+        2 0.00000e+00 2.00000e+00 0.00000e+00
+        3 0.00000e+00 0.00000e+00 3.00000e+00
+
+        >>> assert type(cov) is sandy.CategoryCov
+        """
+        cov = pd.DataFrame(np.diag(std.values),
```
luca-fiorito-11, line 526: maybe it is best to assume that std can be an iterable...not specifically a pandas series.

then we should check: if we have std being a series and passing to `np.diag(std)`, would it still work?
If it works, we don't have to restrict ourselves to a pandas series

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+                           index=std.index, columns=std.index)
+        return cls(cov)
+
+    def GLS_sensitivity(self, S, Vy):
+        """
+        
+
+        Parameters
+        ----------
+        S : `pandas.DataFrame`
```
luca-fiorito-11, line 536: as the above comment...can we generalize and say that this is a 2D iterable and not a dataframe?

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            DESCRIPTION.
+
+        Returns
+        -------
+        TYPE
+            DESCRIPTION.
+
+        """
+        # Posible errors:
+        if S.data.index.values.all() != S.data.columns.values.all():
```
luca-fiorito-11, line 553: this is not tru, the covariance must have identical index and columns...but the sensitivity can be different...also with different sizes, e.g. MXN

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            raise TypeError("Sensitivity matrix has not symmetric indexes")
+        if S.index.values.all() != self.data.index.values.all():
```
luca-fiorito-11, line 555: let's keep it more general if possible and forget about indices. We just write in the docstring that the user should provide consistent size and indices if given

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            raise TypeError('The object indexes and the sensitivity \
+                            indexes do not match')
+        if Vy.index.values.all() != self.data.index.values.all():
```
luca-fiorito-11, line 558: same

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            raise TypeError('The object indexes and the extra information\
+                            indexes do not match')
+        # GLS_sensitivity:
         Vx = self.data.values
-        V = S.T.dot(Vx.dot(S)) + Vy
-        C = Vx - Vx.dot(S.dot(np.linalg.inv(V).dot(S.T.dot(Vx))))
-        return self.__class__(C)
+        M = S.T.dot(Vx.dot(S)) + Vy
```
luca-fiorito-11, line 563: if you want to make sure that `Vy` is a dataframe, independently on what was given, i suggest doing:
```
S_ = pd.Series(S)
Vy_ = pd.DataFrame(Vy)
```
Notice that I now call them `Vy_` and `S_` because I don't want to change the type of the inputs

sandy/core/cov.py
luca-fiorito-11: For S and Vy, explain dimensions with respect to self (using examples like MXN). Otherwise it can be complicated!

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+        2 0.00000e+00 2.00000e+00 0.00000e+00
+        3 0.00000e+00 0.00000e+00 3.00000e+00
+
+        >>> assert type(cov) is sandy.CategoryCov
+        """
+        cov = pd.DataFrame(np.diag(std.values),
+                           index=std.index, columns=std.index)
+        return cls(cov)
+
+    def GLS_sensitivity(self, S, Vy):
```
luca-fiorito-11, line 530: use lowercase letters

luca-fiorito-11: `gls_sensitivity` might by a private method (check what it is)...then it would be `_gls_sensitivity`.

what do you think?

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
         Vx = self.data.values
-        V = S.T.dot(Vx.dot(S)) + Vy
-        C = Vx - Vx.dot(S.dot(np.linalg.inv(V).dot(S.T.dot(Vx))))
-        return self.__class__(C)
+        M = S.T.dot(Vx.dot(S)) + Vy
+        M_inv = sandy.CategoryCov(M).invert()
+        sensitivity = Vx.dot(S).dot(M_inv)
+        return self.__class__(sensitivity)
+
+    def custom_GLS(self, Vx, Vy):
```
luca-fiorito-11, line 568: call it `gls_update`

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            DESCRIPTION.
+
+        Returns
+        -------
+        TYPE
+            DESCRIPTION.
+
+        """
+        # Posible errors:
+        if Vx.index.values.all() != self.data.index.values.all():
```
luca-fiorito-11, line 591: same as above

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            raise TypeError('The object indexes and the sensitivity indexes\
+                             do not match')
+        if Vy.index.values.all() != self.data.index.values.all():
```
luca-fiorito-11, line 594: same as above

sandy/core/cov.py
```diff
@@ -496,11 +496,107 @@ def decompose(self):
...
+            raise TypeError('The object indexes and the extra information\
+                            indexes do not match')
+        # Custom GLS:
+        V_new = Vx + self.data.dot(Vy)
```
luca-fiorito-11, line 598: this is not ok for me.
Shouldn't  it be
```
A = self._gls_sensitivity(S, Vy)
V_new = Vx + A.dot(Vy)
``` 
Then you need S as an input

I thought there was a minus sign...

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: It is okay

sandy/core/cov.py
```diff
@@ -379,29 +379,58 @@ def _reduce_size(self):
         cov_reduced = self.data.loc[nonzero_idxs, nonzero_idxs]
         return nonzero_idxs, cov_reduced
 
-    def _restore_size(cls, nonzero_idxs, cov_reduced, dim):
+    @classmethod
+    def restore_size(cls, nonzero_idxs, cov_reduced, dim):
```
luca-fiorito-11, line 383: whys is it not private anymore?

sandy/core/cov.py
```diff
@@ -514,29 +544,93 @@ def get_covariance(cls, std):
...
+        >>> assert type(sandy.CategoryCov.from_var([1, 2, 3])) is sandy.CategoryCov
+        """
+        if hasattr(var, "__len__") and type(var) != str:
+            if type(var) == pd.Series:
+                cov = pd.DataFrame(np.diag(var.values),
+                                   index=var.index, columns=var.index)
+            else:
+                cov = pd.DataFrame(np.diag(var))
+        else:
+            raise TypeError("Variance vector is not valid")
```
luca-fiorito-11, line 573: this is not needed

sandy/core/cov.py
```diff
@@ -514,29 +544,93 @@ def get_covariance(cls, std):
...
+        0 1.00000e+00 0.00000e+00 0.00000e+00
+        1 0.00000e+00 2.00000e+00 0.00000e+00
+        2 0.00000e+00 0.00000e+00 3.00000e+00
+
+        >>> assert type(S) is sandy.CategoryCov
+        >>> assert type(sandy.CategoryCov.from_var([1, 2, 3])) is sandy.CategoryCov
+        """
+        if hasattr(var, "__len__") and type(var) != str:
+            if type(var) == pd.Series:
+                cov = pd.DataFrame(np.diag(var.values),
```
luca-fiorito-11, line 568: are you sure this doesn't work with `np.diag(var)`?

sandy/core/cov.py
```diff
@@ -514,29 +544,93 @@ def get_covariance(cls, std):
...
+
+    @classmethod
+    def from_stdev(cls, std):
+        """
+        Construct the covariance matrix from the standard deviation vector.
+
+        Parameters
+        ----------
+        std : `pandas.Series`
+            Variance vector.
```
luca-fiorito-11, line 584: standard deviations

sandy/core/cov.py
```diff
@@ -514,29 +544,93 @@ def get_covariance(cls, std):
...
-                           index=std.index, columns=std.index)
+        if hasattr(std, "__len__") and type(std) != str:
+            if type(std) == pd.Series:
+                cov = pd.DataFrame(np.diag(std.values*std.values),
+                                   index=std.index, columns=std.index)
+            else:
+                std_ = np.array(std)
+                cov = pd.DataFrame(np.diag(std_*std_))
+        else:
+            raise TypeError("Standart deviation vector is not valid")
```
luca-fiorito-11, line 621: same as above

sandy/core/cov.py
```diff
@@ -514,29 +544,93 @@ def get_covariance(cls, std):
...
         Parameters
         ----------
-        S : `pandas.DataFrame`
-            DESCRIPTION.
-        Vy : `pandas.Series`
-            DESCRIPTION.
+        S : 2D iterable
+            Sensitivity square matrix (MXM).
+        Vy : 1D iterable
+            Extra Covariance vector (MX1).
```
luca-fiorito-11, line 633: missing threshold

sandy/core/cov.py
```diff
@@ -549,53 +643,99 @@ def GLS_sensitivity(self, S, Vy):
             DESCRIPTION.
 
         """
+        # Check the data entered
+        if hasattr(S, "__len__") and type(S) != str:
+            if type(S) != pd.DataFrame:
+                S_ = pd.Dataframe(S)
```
luca-fiorito-11, line 649: this works also for dataframes

sandy/core/cov.py
```diff
@@ -549,53 +643,99 @@ def GLS_sensitivity(self, S, Vy):
             DESCRIPTION.
 
         """
+        # Check the data entered
+        if hasattr(S, "__len__") and type(S) != str:
+            if type(S) != pd.DataFrame:
+                S_ = pd.Dataframe(S)
+            else:
+                S_ = S
+        else:
```
luca-fiorito-11, line 652: not needed

sandy/core/cov.py
```diff
@@ -549,53 +643,99 @@ def GLS_sensitivity(self, S, Vy):
...
+            if type(S) != pd.DataFrame:
+                S_ = pd.Dataframe(S)
+            else:
+                S_ = S
+        else:
+            raise TypeError('Introduced sensitivity is not valid')
+
+        if hasattr(Vy, "__len__") and type(Vy) != str:
+            if type(Vy) != pd.Series:
+                Vy_ = pd.Series(Vy)
```
luca-fiorito-11, line 657: this works also for series

sandy/core/cov.py
```diff
@@ -549,53 +643,99 @@ def GLS_sensitivity(self, S, Vy):
...
+            else:
+                Vy_ = Vy
+        else:
+            raise TypeError('Introduced extra covariance vector is not valid')
```
luca-fiorito-11, line 661: not needed

sandy/core/cov.py
```diff
@@ -549,53 +643,99 @@ def GLS_sensitivity(self, S, Vy):
...
+            raise TypeError('The object columns and the sensitivity \
+                            columns do not match')
+        if not S_.columns.intersection(Vy_.index).any():
+            raise TypeError('Object indexes and the extra information\
                             indexes do not match')
-        if Vy.index.values.all() != self.data.index.values.all():
-            raise TypeError('The object indexes and the extra information\
-                            indexes do not match')
+        elif len(S_.columns.intersection(Vy_.index)) < len(S_.columns):
+            Vy_ = Vy_.reindex(S_.columns.index, fill_value=0)
```
luca-fiorito-11, line 674: not sure we need all these things

luca-fiorito-11 (changes requested):
I still need to finish something

sandy/core/cov.py
AitorBengoechea: My mistake

sandy/core/cov.py
AitorBengoechea: Works

AitorBengoechea: It cant be done: with np.diag, the previous index are lost

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+        M = S_.T.dot(Vx.dot(S_)) + Vy_
+        M_inv = sandy.CategoryCov(M).invert()
+        sensitivity = Vx.dot(S_).dot(M_inv).dot(S_.T)
+        if threshold is not None:
+            sensitivity[sensitivity < threshold] = 0
+        return self.__class__(sensitivity)
+
+    def gls_update(self, Vy, S, delta, threshold=None):
+        """
+
```
luca-fiorito-11, line 650: missing description

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+            Thereshold to avoid numerical fluctuations. The default is None.
+
+        Returns
+        -------
+        TYPE
+            DESCRIPTION.
+
+        """
+        # Check the data entered
+        S_ = pd.Dataframe(S)
```
luca-fiorito-11, line 670: this is not needed...it is already done in `self._gls_sensitivity`

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+            Sensitivity square matrix (MXM).
+        delta : 1D iterable
+            Perturbed vector minus non perturbed vector.
+        threshold : `int`, optional
+            Thereshold to avoid numerical fluctuations. The default is None.
+
+        Returns
+        -------
+        TYPE
+            DESCRIPTION.
```
luca-fiorito-11, line 666: missing

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+
```
luca-fiorito-11, line 667: can we define a very simple test...maybe a 2x2 matrix for the prior, a 2x1 for the sensitivity and a 1x1 for the new info

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+        -------
+        `CategoryCov`
+            `CategoryCov` object to which we have applied sandwich
+            formula for a given pd.Series
+
+        Examples
+        --------
+        >>> S = np.diag(np.array([1, 2, 3]))
+        >>> var = pd.Series([1, 2, 3])
+        >>> cov = sandy.CategoryCov(S)
```
luca-fiorito-11, line 701: instead of doing this you can call `from_var`

sandy/core/cov.py
luca-fiorito-11: maybe we should remove this...again we expect that the users know what they are doing.
just specify in the docstring!

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
+            raise TypeError("The indices of the sensitivity array \
+                              and of the CategoryCov are not the same")
         C = self.data.values
-        Si_ = Si.reshape(self.size, -1)
-        if Sj is None:
-            Sj_ = Si_
-        else:
-            Sj_ = Sj.reshape(self.size, -1)
-        return Si_.T.dot(C.dot(Sj_))
+        s = S.values
```
luca-fiorito-11, line 723: i would do `pd.Series(S)`

sandy/core/cov.py
luca-fiorito-11: please, fix this

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -393,20 +532,197 @@ def decompose(self):
...
         Vx = self.data.values
-        V = S.T.dot(Vx.dot(S)) + Vy
-        C = Vx - Vx.dot(S.dot(np.linalg.inv(V).dot(S.T.dot(Vx))))
-        return self.__class__(C)
+        M = S_.T.dot(Vx.dot(S_)) + Vy_
+        M_inv = sandy.CategoryCov(M).invert()
+        sensitivity = Vx.dot(S_).dot(M_inv).dot(S_.T)
+        if threshold is not None:
+            sensitivity[sensitivity < threshold] = 0
+        return self.__class__(sensitivity)
```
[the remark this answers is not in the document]
AitorBengoechea: It trues. Thanks, i was block on that

AitorBengoechea:
Only the tests for GLS are missing. I have removed the symmetry issue because if it is there, all methods that update CategoryCov fail.

sandy/core/cov.py
```diff
@@ -258,12 +259,6 @@ def data(self):
     @data.setter
     def data(self, data):
         self._data = data.astype(float)
```
luca-fiorito-11: Maybe it should be `self._data = pd.DataFrame(data, dtype=float)`.
Then your input can also be a numpy matrix.

sandy/core/cov.py
```diff
@@ -468,10 +463,10 @@ def invert(self):
         unit = np.identity(len(M_reduce))
         M_reduce_inv = splu(csc_matrix(M_reduce)).solve(unit)
         data = CategoryCov._restore_size(M_nonzero_idxs, M_reduce_inv,
-                                        len(self.data))
+                                         len(self.data))
         M_inv = pd.DataFrame(data.data,
                              index=index, columns=columns)
-        return self.__class__(M_inv)
+        return M_inv
```
luca-fiorito-11, line 469: Why not `self.__class__`?

sandy/core/cov.py
```diff
@@ -643,11 +638,11 @@ def _gls_sensitivity(self, S, Vy, threshold=None):
         sensitivity = Vx.dot(S_).dot(M_inv).dot(S_.T)
         if threshold is not None:
             sensitivity[sensitivity < threshold] = 0
-        return self.__class__(sensitivity)
+        return pd.DataFrame(sensitivity, index=S_.index, columns=S_.columns)
 
     def gls_update(self, Vy, S, delta, threshold=None):
         """
-        
+        Perform GlS method for a given variance and sensitivity.
```
luca-fiorito-11, line 645: Update, not method

sandy/core/cov.py
```diff
@@ -662,18 +657,13 @@ def gls_update(self, Vy, S, delta, threshold=None):
...
         """
-        # Check the data entered
-        S_ = pd.Dataframe(S)
-        Vy_ = pd.Series(Vy)
-
-        # Custom GLS:
         Vx = self.data
-        A = self._gls_sensitivity(S_, Vy_)
-        V_new = Vx - A.data.dot(Vx)
+        A = self._gls_sensitivity(S, Vy)
```
luca-fiorito-11, line 665: Why don't you do anymore `pd.DataFrame(S)`?

sandy/core/cov.py
luca-fiorito-11: This is still here

sandy/core/cov.py
```diff
@@ -258,12 +259,6 @@ def data(self):
     @data.setter
     def data(self, data):
         self._data = data.astype(float)
-        if not len(data.shape) == 2 and data.shape[0] == data.shape[1]:
-            raise sandy.Error("covariance matrix must have two dimensions")
-        if not np.allclose(data.values, data.values.T):
-            raise sandy.Error("covariance matrix must be symmetric")
-        if (np.diag(data) < 0).any():
```
luca-fiorito-11: You can delete the checks, for the moment, but write in the docstring that they must be implemented (and which ones)

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: Because it is done in _gls_sensitivity for the two variables

AitorBengoechea:
@luca-fiorito-11  finished the work. Waiting for the corrections.

sandy/core/cov.py
```diff
@@ -393,20 +534,216 @@ def decompose(self):
...
+                     0            1
+        0  7.42857e-01 -3.14286e-01
+        1 -3.14286e-01  1.71429e-01
+        """
+        index, columns = self.data.index, self.data.columns
+        Vx = self.data.values
+        A = self._gls_sensitivity(S, Vy).values
+        V_new = Vx - A.dot(Vx)
+        V_new = pd.DataFrame(V_new, index=index, columns=columns)
+        if threshold is not None:
```
luca-fiorito-11, line 700: this threshold is weird...I thought it was supposed to be passed to `_gls_sensitivity`.
Where do we need the threshold? on the variance or on the sensitivity?

luca-fiorito-11 (commented):
one last comment!

luca-fiorito-11 (approved):

PR #93, opened 2021-11-29
DLAMBDA added to mf8

EnricaBelfiore:


luca-fiorito-11:
It looks like the entire file was modified...not only the section about uncertainties.
Can you check why?
Try with a `git diff`
Maybe it's that your file changed the line breaks from windows to Linux or vice-versa.
Try running `dos2unix`

luca-fiorito-11:
Also, can you write an example to check what you changed?

PR #94, opened 2021-11-30
Uncertainty added

EnricaBelfiore:
I don't like my example because I think that the user will never use it. Do you want an example that shows what I changed (the uncertainty)?


sandy/sections/mf8.py
```diff
@@ -288,7 +294,9 @@ def _read_rdd(tape, mat):
             # radiation types
             "DE": L.B[1::2],
             # decay constant in 1/s, 0 if stable
-            "LAMBDA": math.log(2.0)/L.C1 if L.C1 else 0
+            "LAMBDA": math.log(2.0)/L.C1 if L.C1 else 0,
+            # uncertainty on decay constant 
+            "DLAMBDA": math.log(2.0)*L.C2/pow(L.C1,2) if L.C1 else 0
```
luca-fiorito-11, line 299: PEP8

sandy/sections/mf8.py
```diff
@@ -288,7 +294,9 @@ def _read_rdd(tape, mat):
             # radiation types
             "DE": L.B[1::2],
             # decay constant in 1/s, 0 if stable
-            "LAMBDA": math.log(2.0)/L.C1 if L.C1 else 0
+            "LAMBDA": math.log(2.0)/L.C1 if L.C1 else 0,
```
luca-fiorito-11, line 297: PEP8

sandy/sections/mf8.py
luca-fiorito-11: can you use `L.C1**2` instead of `pow(L.C1,2)`

Upload/Perturbation_decay_constant.ipynb
```diff
@@ -0,0 +1,270 @@
+{
```
luca-fiorito-11, line 1: this file must be removed

luca-fiorito-11 (changes requested):
btw, did you run pytest?

sandy/sections/mf8.py
luca-fiorito-11: this is still open

sandy/sections/mf8.py
```diff
@@ -288,7 +295,9 @@ def _read_rdd(tape, mat):
...
+            # uncertainty on decay constant
+            "DLAMBDA": math.log(2.0)*L.C2/L.C1**2 if L.C1 else 0
```
luca-fiorito-11, line 300: PEP8

luca-fiorito-11 (changes requested):

luca-fiorito-11:
You MUST run pytest **every time**! If you don't do it I cannot trust your changes.
There's no way your test passed!

please, bring back your previous changes, i.e., `"LAMBDA": math.log(2.0) / L.C1 if L.C1 else 0,`

sandy/sections/mf8.py
luca-fiorito-11: with this I mean that you should write `math.log(2.0) / L.C1` instead of `math.log(2.0)/L.C1`

sandy/sections/mf8.py
```diff
@@ -288,8 +293,11 @@ def _read_rdd(tape, mat):
             # radiation types
             "DE": L.B[1::2],
             # decay constant in 1/s, 0 if stable
-            "LAMBDA": math.log(2.0)/L.C1 if L.C1 else 0
+            "LAMBDA": math.log(2.0) / L.C1 if L.C1 else 0,
+            # uncertainty on decay constant
+            "DLAMBDA": math.log(2.0)*L.C2 / L.C1**2 if L.C1 else 0
```
luca-fiorito-11, line 298: `math.log(2.0) * L.C2`, not `math.log(2.0)*L.C2`

sandy/sections/mf8.py
```diff
@@ -248,6 +248,11 @@ def _read_rdd(tape, mat):
     'DRICK': 0.0,
     'RICL': 0.0,
     'DRICL': 0.0}
+
```
luca-fiorito-11, line 251: can you put back the previous test? that test would have failed in your previous commit...so, it is better to keep it!

luca-fiorito-11 (changes requested):

sandy/sections/mf8.py
EnricaBelfiore: I did a test with the previous example and I did not find errors, anyway I rechanged it

luca-fiorito-11: But now, why did you delete the other test?

luca-fiorito-11 (approved):

PR #96, opened 2021-12-01
gls_cov_update

AitorBengoechea:


AitorBengoechea:
For the moment, check this method and if it is correct, i will upload the test

AitorBengoechea:
Both methods were implemented. After code revision, I will upload the test. My results (written ones) and the results obtained with the methods are exactly the same.

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
         new_data.loc[mask, 'FY'] = fy_calc.values
         return self.__class__(new_data)
 
+    def gls_cov_update(self, zam, e, decay_data, Vy=None, kind='cumulative',
+                       threshold=None, index_columns=None):
+        """
+        IFY covariance matrix performed by GLS for a given energy and zam.
```
luca-fiorito-11, line 363: `Update the prior IFY covariance matrix using the GLS technique described in https://doi.org/10.1016/j.anucene.2015.10.027`

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            account. The default is None.
+        index_columns : `str`, optional
+            Keyword to determine if the sensitivity is the transposed one.
+            The default is None.
+
+        Returns
+        -------
+        `panda.DataFrame`
+            IFY covariance matrix performed by GLS for a given energy and zam.
+
```
luca-fiorito-11, line 392:
```
Notes
-----
..note:: only option `kind='cumulative'` is implemented
```

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        # Create the Cov object:
+        DIFY = sandy.CategoryCov.from_var(DIFY)
+        if Vy is None:
+            Vy = sandy.CategoryCov.from_var(DCFY).data
+        # Fix the index:
+        S = S.loc[DIFY.index, DIFY.index].fillna(0)
+        return DIFY.gls_update(S, Vy, threshold=threshold).data
+
+    def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
+                   threshold=None, index_columns=None):
```
luca-fiorito-11, line 411: i don't get the index column...the code should know whether to transpose or not, not the user

sandy/fy.py
```diff
@@ -524,6 +633,41 @@ def to_hdf5(self, file, library):
...
+    Returns
+    -------
+    S : `pandas.DataFrame`
+        The prior sensitivity for performing GLS.
+
+    """
+    if kind == 'cumulative':
+        S = decay_data.get_qmatrix()
+    elif kind == 'chain':
+        S = decay_data.get_decay_chains()
```
luca-fiorito-11, line 663: this is not the sensitivity matrix for chain fission yields.

sandy/fy.py
```diff
@@ -524,6 +633,41 @@ def to_hdf5(self, file, library):
...
+    else:
+        raise TypeError('The kind introduced is not valid')
```
luca-fiorito-11, line 665: it is more a `ValueError` than a `TypeError`

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            applied.
+        e : `float`
+            Energy to which calculations are to be applied.
+        decay_data : `sandy.DecayData`
+            Radioactive nuclide data for several isotopes.
+        xp : 1D iterable
+            New CFY information vector.
+        Vy : 2D iterable, optional
+            Extra Covariance matrix (MXM). The default is None, so the method
+            uses the variance of the CFY for a given zam and e.
```
luca-fiorito-11, line 428: if Vy is optional, then also xp is optional. Otherwise none of them is!
For the moment I would like them not to be optional keywords. In theory the default option should be chain yields.

luca-fiorito-11: why do you call it xp (it sounds like x prior, which is not the case). I guess y should be a better name, to be consistent with Vy

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        # Divide the data:
+        fy_type = fy_data.groupby('MT')
+        IFY, CFY = [fy_type.get_group(x)['FY'] for x in fy_type.groups]
+        DIFY, DCFY = [fy_type.get_group(x)['DFY'] for x in fy_type.groups]
+        # Put the data in a appropiate format:
+        DIFY = sandy.CategoryCov.from_var(DIFY).data
+        if Vy is None:
+            Vy = sandy.CategoryCov.from_var(DCFY).data
+        # Fix the index:
+        S = S.loc[IFY.index, IFY.index].fillna(0)
```
luca-fiorito-11, line 462: this is not True, once again, sensitivities might have different indices (you will see it with chain yields, where indices are not nuclides but mass numbers)

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        zam : `int`
+            ZAM number of the material to which calculations are to be
+            applied.
+        e : `float`
+            Energy to which calculations are to be applied.
+        decay_data : `sandy.DecayData`
+            Radioactive nuclide data for several isotopes.
+        Vy : 2D iterable, optional
+            Extra Covariance matrix (MXM). The default is None, so the method
+            uses the variance of the CFY for a given zam and e.
```
luca-fiorito-11, line 376: For the moment I would like Vy not to be an optional keywords. In theory the default option should be chain yields.

sandy/fy.py
```diff
@@ -357,6 +357,115 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+                      .set_index('ZAP')[['MT', 'DFY']]
+        # Divide the data type:
+        fy_type = fy_data.groupby('MT')
+        DIFY, DCFY = [fy_type.get_group(x)['DFY'] for x in fy_type.groups]
+        # Create the Cov object:
+        DIFY = sandy.CategoryCov.from_var(DIFY)
+        if Vy is None:
+            Vy = sandy.CategoryCov.from_var(DCFY).data
+        # Fix the index:
+        S = S.loc[DIFY.index, DIFY.index].fillna(0)
```
luca-fiorito-11, line 407: this is not True, once again, sensitivities might have different indices (you will see it with chain yields, where indices are not nuclides but mass numbers)

luca-fiorito-11 (changes requested):

sandy/fy.py
AitorBengoechea: I put two optional index-related arguments in _gls_setup. I put them in this way since what we get from decay data is a DataFrame and we might want to reduce the size of the S. 
In the specific case of CFY and IFY, qmatrix >(CFY and IFY), therefore, by using these two arguments, we reduce the size of qmatrix by selecting only the nucleus that are related to the IFY we are analysing.

luca-fiorito-11: I understand it for the first dimension of the sensitivity matrix, but not for the second. If you work with chain fission yields you don't have the index of `DIFY`. Then the second index (the one with `??`) will not work,

`S.loc[DIFY.index, ??DIFY.index??]`

luca-fiorito-11 (changes requested):

sandy/fy.py
AitorBengoechea: The sensitivity matrix for chain fission yields are you meaning the matrix obtained with the method get_transition_matrix()?

sandy/fy.py
```diff
@@ -658,13 +645,17 @@ def _gls_setup(decay_data, kind, index_columns):
 
     """
     if kind == 'cumulative':
-        S = decay_data.get_qmatrix()
+        S = decay_data.get_qmatrix().T
     elif kind == 'chain':
-        S = decay_data.get_decay_chains()
+        S = decay_data.get_transition_matrix()
```
luca-fiorito-11, line 650: It is not `get_transition_matrix`. We have to develop it! Don't implement it for the moment!

sandy/fy.py
luca-fiorito-11: Not resolved!

sandy/fy.py
```diff
@@ -371,44 +372,37 @@ def gls_cov_update(self, zam, e, decay_data, Vy=None, kind='cumulative',
...
-        S = _gls_setup(decay_data, kind, index_columns=None)
-        # Filter FY data:
-        fy_data = self.filter_by('ZAM', zam)\
-                      .filter_by("E", e).data\
-                      .set_index('ZAP')[['MT', 'DFY']]
         # Divide the data type:
-        fy_type = fy_data.groupby('MT')
-        DIFY, DCFY = [fy_type.get_group(x)['DFY'] for x in fy_type.groups]
+        mask = (self.data.ZAM == zam) & (self.data.MT == 454) & (self.data.E == e)
+        DIFY, index = self.data.loc[mask, 'DFY'], self.data.loc[mask].set_index('ZAP').index
```
luca-fiorito-11, line 397: it would be the same (arguably better) if you do:
```
DIFY = self.data.loc[mask, 'DFY']
index = self.data.loc[mask].ZAP.values

File: sandy/fy.py [hunk 12]
```diff
@@ -423,9 +417,8 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
             Radioactive nuclide data for several isotopes.
         xp : 1D iterable
             New CFY information vector.
```
luca-fiorito-11, line 419: variable name must be updated

sandy/fy.py
```diff
@@ -423,9 +417,8 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
...
-        Vy : 2D iterable, optional
-            Extra Covariance matrix (MXM). The default is None, so the method
-            uses the variance of the CFY for a given zam and e.
+        Vy : 2D iterable
+            Extra Covariance matrix (MXM).
         kind : `str`, optional
             Keyword for obtaining prior sensitivity. The
```
luca-fiorito-11, line 423: it's just sensitivity, not prior sensitivity

sandy/fy.py
```diff
@@ -445,24 +438,21 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
...
         """
         data = self.data
-        S = _gls_setup(decay_data, kind, index_columns)
         # Filter FY data:
         fy_data = self.filter_by('ZAM', zam)\
                       .filter_by("E", e).data\
                       .set_index('ZAP')[['MT', 'FY', 'DFY']]
         # Divide the data:
         fy_type = fy_data.groupby('MT')
         IFY, CFY = [fy_type.get_group(x)['FY'] for x in fy_type.groups]
```
luca-fiorito-11, line 447: here you are assuming that `for x in fy_type.groups` returns the MT numbers in increasing order.
I am not sure this is the case!
Maybe it just gives them back according to what MT number was found first.
I would avoid complexity by simplifying the format...something like

```
IFY = fy_data.query('MT==454').FY
CFY = fy_data.query('MT==459).FY
```

sandy/fy.py
```diff
@@ -445,24 +438,21 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
...
-        DIFY, DCFY = [fy_type.get_group(x)['DFY'] for x in fy_type.groups]
+        mask = (data.ZAM == zam) & (data.MT == 454) & (data.E == e)
+        DIFY = data.loc[mask, 'DFY']
```
luca-fiorito-11, line 449: here you are filtering again...but everything was already filtered in `fy_data`.

Couldn't you do `DIFY = fy_data.query('MT==459).DFY` ?

sandy/fy.py
luca-fiorito-11: Ah, PEP20 says that *sparse is better than dense*, then I would not do two operations on the same line

sandy/fy.py
```diff
@@ -445,24 +438,21 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
...
+        # Fix the S with the correct dimension:
+        S = _gls_setup(decay_data, kind, index=IFY.index, columns=IFY.index)
```
luca-fiorito-11, line 451: Here we are again! the sensitivity is MxN, not MxM. Then index and columns in general are not the same! It looks this concept is not clear!

in the linear model `y=S*x`, where `y is Mx1` and `x is Nx1`, `S` must be `MxN`! nobody says that y and x have the same dimensions (and the same index)!

Another examples, i consider a fuel decay heat model that depends on the independent yields, something like `DH=f(IFY)` where `DY` is a scalar value in Watt, i.e., the fuel total decay heat. Do y and x have the same index? NO!

sandy/fy.py
```diff
@@ -445,24 +438,21 @@ def gls_update(self, zam, e, decay_data, xp, Vy=None, kind='cumulative',
...
         # Put the data in a appropiate format:
         DIFY = sandy.CategoryCov.from_var(DIFY).data
-        if Vy is None:
-            Vy = sandy.CategoryCov.from_var(DCFY).data
-        # Fix the index:
-        S = S.loc[IFY.index, IFY.index].fillna(0)
         # Perform GLS:
-        IFY_new = sandy.gls_update(IFY, S, DIFY, Vy, CFY, xp, threshold=threshold)
-        mask = (data.ZAM == zam) & (data.MT == 454) & (data.E == e)
+        IFY_new = sandy.gls_update(IFY, S, DIFY, Vy, CFY, y, threshold=threshold)
```
luca-fiorito-11, line 455: I see here that there is something wrong with `sandy.gls_update`...`CFY` shouldn't be given. It should be internally calculated by as a solution of the linearized model `y=S x`

luca-fiorito-11 (changes requested):

PR #98, opened 2021-12-01
get_decayconstant

EnricaBelfiore:


sandy/decay.py
```diff
@@ -114,6 +114,40 @@ def get_halflives(self):
...
+                   DLAMBDA      LAMBDA
+           922350 2.21715e-20 3.12085e-17
+        """
+
+        decay_constant = {zam: {
+             "LAMBDA": dic['decay_constant'],
+             "DLAMBDA": dic['decay_constant_uncertainty'],
+             } for zam, dic in self.data.items()}
+
+        df = pd.DataFrame(decay_constant)
```
luca-fiorito-11, line 145: I would do
```
df = pd.DataFrame(decay_constant).T
if with_uncertainty:
            return df
else:
            return df.LAMBDA
```
I am not sure `iat` is the best for this case

luca-fiorito-11: But if you do this you need to say in the docstring that you get a series in output if with_uncertainty is false

sandy/decay.py
```diff
@@ -333,6 +367,7 @@ def get_transition_matrix(self):
               .fillna(0)
         return T.reindex(T.columns.values, fill_value=0.0)
 
+
```
luca-fiorito-11, line 370: PEP8, Remove this line

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: Let me know if what I wrote is the correct way to explain the two output options

sandy/decay.py
```diff
@@ -136,17 +137,16 @@ def get_decayconstant(self, with_uncertainty=True):
                    DLAMBDA      LAMBDA
            922350 2.21715e-20 3.12085e-17
```
luca-fiorito-11, line 138: you should also test `with_uncertainty=False`

sandy/decay.py
```diff
@@ -114,6 +114,40 @@ def get_halflives(self):
         series.index.name = "ZAM"
         return series
 
+    def get_decayconstant(self, with_uncertainty=True):
+        """
+        Extract decay constant and its uncertainty into dataframe.
+
+        Parameters
+        ----------
+        with_uncertainty : `bool`, optional, default is 'True'
```
luca-fiorito-11, line 123: you did not describe this

luca-fiorito-11:
@EnricaBelfiore, I merged the branch, although I'd like one more test

PR #99, opened 2021-12-02
custom perturbation

EnricaBelfiore:


sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        --------
+        >>> endf6 = sandy.get_endf6_file("jeff_33",'decay',[922350, 922330, 942390, 942320, 922350])
+        >>> rdd = sandy.DecayData.from_endf6(endf6)
+        >>> tape =sandy.DecayData.custom_perturbation_decayconstant(rdd, 0.05, [922350, 922330])
+        >>> print(tape)
+        922350   3.27689e-17
+        922330   1.44821e-13
+        Name: LAMBDA, dtype: float64
+        """
+        tape = sandy.DecayData.get_decayconstant(self)
```
luca-fiorito-11, line 183: It's weird that you call it `tape`, which is generally used for endf6 files.
This is just a pd.dataframe.

Maybe just call it `dc`.

And since you don't use the uncertainty just do:
`dc = self.get_decayconstant().LAMBDA`.

BTW, I don't get your way of calling self

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        if zam != 'all':
```
luca-fiorito-11, line 184: I don't like the 'all' option. We are never in the situation of perturbing all decay constants.
I would either have it not ad default, or even better remove it....it will simplify the method

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
 
+    def custom_perturbation_decayconstant(self, pert, zam="all"):
+        """
+        Apply a custom perturbation to the decay constant
+        Parameters
+        ----------
+        zam : `int` or `list` or `all`, default is `all`
+            ZAM number of the material to which perturbation is to be
+            applied.
+        pert : 'float'
```
luca-fiorito-11, line 167: `float`

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+            Perturation coeffcient as ratio value.
```
luca-fiorito-11, line 168: Typo

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        Returns
+        -------
+        `sandy.decay.DecayData`
+            DacayData instance with given value ZAM and decay constant perturbed.
+        Examples
+        --------
+        >>> endf6 = sandy.get_endf6_file("jeff_33",'decay',[922350, 922330, 942390, 942320, 922350])
```
luca-fiorito-11, line 175: Missing spaces

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        ----------
+        zam : `int` or `list` or `all`, default is `all`
+            ZAM number of the material to which perturbation is to be
+            applied.
+        pert : 'float'
+            Perturation coeffcient as ratio value.
+        Returns
+        -------
+        `sandy.decay.DecayData`
+            DacayData instance with given value ZAM and decay constant perturbed.
```
luca-fiorito-11, line 172: Missing empty line

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        Name: LAMBDA, dtype: float64
+        """
+        tape = sandy.DecayData.get_decayconstant(self)
+        if zam != 'all':
+            for z in zam:
+                tape.LAMBDA[z] = tape.LAMBDA[z] * (1 + pert)
+            return self.__class__(tape.LAMBDA[zam])
+        else:
+            tape.LAMBDA = tape.LAMBDA * (1 + pert)
+            return self.__class__(tape.LAMBDA)
```
luca-fiorito-11, line 190: The docstring says you should return a DecayData...I am not sure what you return, but it's wrong

sandy/decay.py
```diff
@@ -148,6 +156,39 @@ def get_decayconstant(self, with_uncertainty=True):
         else:
             return df.LAMBDA
 
+    def custom_perturbation_decayconstant(self, pert, zam="all"):
+        """
+        Apply a custom perturbation to the decay constant
+        Parameters
```
luca-fiorito-11, line 162: Missing empty line

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: Do you prefer a pandas Series as outout? I printed the type of the output that I obtained and it said <class 'sandy.decay.DecayData'>, for this reason I wrote that in the docstring but I understand that I can do it differently

sandy/decay.py
EnricaBelfiore: I wrote the 'all' option because I supposed that the user can have different isotopes and can decide to perturb a specific one (I didn't mean all the decay constants from the decay data) but surely  he can use a filter to do this. If you think that this option is not valid I will remove it.

sandy/decay.py
luca-fiorito-11: No, as we mentioned a few times already, you should be able to to rewrite the DecayData instance with perturbed decay constants into an Endf6 instance and then to file.
At the moment you cannot do it because, despite you return a DecayData instance, it does not contain the right data...then all DecayData methods will fail, including `to_endf6`.

returning a pandas Series would just lead you even further away from the problem.

If you have problems in seeing the bigger picture, try and develop more steps on a notebook

sandy/decay.py
```diff
@@ -148,6 +156,32 @@ def get_decayconstant(self, with_uncertainty=True):
...
+        --------
+        >>> endf6 = sandy.get_endf6_file("jeff_33", 'decay', [922350, 922330, 922350])
+        >>> rdd = sandy.DecayData.from_endf6(endf6)
+        >>> dc_pert = sandy.DecayData.custom_perturbation_decayconstant(rdd, 0.05)
+        >>> print(dc_pert)
+        {922330: 1.4482087325525418e-13, 922350: 3.2768932273817556e-17}
+        """
+        dc = self.get_decayconstant().LAMBDA
+        dc = dc * (1 + pert)
+        return self.__class__(dc.to_dict())
```
luca-fiorito-11, line 183: this cannot work!!!...you see it in your example.
When you `print(dc_pert)` - which is supposed to be a `DecayData` instance - you show the perturbed decay constants. Now, where are the decay energies? the spin? the half-life?....if you don't have them then your `DecayData` is not correct.

in theory we should add some `if` clauses in the `data.setter` of `DecayData` to make sure you are initializing correctly the instance.
@AitorBengoechea , what do we do for FYs?

luca-fiorito-11 (changes requested):

sandy/decay.py
```diff
@@ -148,6 +156,54 @@ def get_decayconstant(self, with_uncertainty=True):
...
+                  'decay_modes': {'40x0': {'decay_products': {902310: 1.0, 20040: 1.0},
+                    'branching_ratio': 1.0,
+                    'branching_ratio_uncertainty': 0.0001},
+                   '60x0': {'decay_products': {},
+                    'branching_ratio': 7.2e-11,
+                    'branching_ratio_uncertainty': 2.1e-11}}}}
+        """
+        dc = self.get_decayconstant().LAMBDA
+        dc = dc * (1 + pert)
+        pert_dc = self.data
```
luca-fiorito-11, line 202: This should be `self.data.copy()`...or something like that.
Otherwise you will also be changing your input `DecayData` instance.

Please, check this carefully on internet

sandy/decay.py
```diff
@@ -148,6 +156,54 @@ def get_decayconstant(self, with_uncertainty=True):
...
+                   'total': 163320.33067324167},
+                  'decay_modes': {'40x0': {'decay_products': {902310: 1.0, 20040: 1.0},
+                    'branching_ratio': 1.0,
+                    'branching_ratio_uncertainty': 0.0001},
+                   '60x0': {'decay_products': {},
+                    'branching_ratio': 7.2e-11,
+                    'branching_ratio_uncertainty': 2.1e-11}}}}
+        """
+        dc = self.get_decayconstant().LAMBDA
+        dc = dc * (1 + pert)
```
luca-fiorito-11, line 201: We are missing the ZAM here. As mentioned before we don't want to perturb all decay constants.

luca-fiorito-11 (changes requested):

sandy/decay.py
luca-fiorito-11: @EnricaBelfiore , is it clear why we want to perturb only one ZAM?

EnricaBelfiore: Not so clear I think. It cannot be 'all', but why not a list?

EnricaBelfiore: Sorry, stupid question. I wrote a notebook and I get it.

luca-fiorito-11: If it were a list of zam, then I would expect a list of perturbations (we don't want to apply the same perturbation to all decay constants, right?! Imagine when we do random sampling).
But then why not have the function apply the perturbation to only one decay constant and the loop around it (maybe with another function). This follows the philosophy of having functions performing one and only one task.

sandy/decay.py
```diff
@@ -197,12 +200,9 @@ def custom_perturbation_decayconstant(self, pert):
...
         """
-        dc = self.get_decayconstant().LAMBDA
-        dc = dc * (1 + pert)
-        pert_dc = self.data
-        for key in self.data:
-            pert_dc[key]['decay_constant'] = dc[key]
-        return self.__class__(pert_dc)
+        pert_dc = self.data[zam].copy()
+        pert_dc['decay_constant'] = pert_dc['decay_constant'] * (1 + pert)
+        return self.__class__({zam: pert_dc})
```
luca-fiorito-11, line 205: I like this better...but here you return in the DecayData output only the ZAM to which you appiedato the perturbation, while the others are lost.

For instance, in the example you lose the info on U238.
Is this the behavior you want? I think it makes more sense to keep all nuclides....what do you think?

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: I thought that the output could only show the zam with the updated decay constant but you are right, it is better to report all the zam otherwise to see and use them I have to recall the input file. I fixed it and also imported the copy module to use deepcopy to edit only the copy

luca-fiorito-11 (approved):

PR #100, opened 2021-12-03
Fy object

AitorBengoechea:


sandy/fy.py
```diff
@@ -350,13 +352,151 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            account. The default is None.
+        index_columns : `str`, optional
+            Keyword to determine if the sensitivity is the transposed one.
+            The default is None.
+
+        Returns
+        -------
+        `FY`
+            IFY updated with GLS for a given zam, energy, decay_data and
+            new CFY information.
```
luca-fiorito-11, line 478: not `CFY`...just new information

sandy/fy.py
```diff
@@ -350,13 +352,151 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        fy_data = data.filter_by('ZAM', zam)\
+                      .filter_by("E", e).data\
+                      .set_index('ZAP')[['MT', 'FY', 'DFY']]
+        # Divide the data:
+        IFY = fy_data.query('MT==454').FY
+        mask = (data.data.ZAM == zam) & (data.data.MT == 454) & (data.data.E == e)
+        DIFY = fy_data.query('MT==459').DFY
+        # Find the GLS varibles:
+        S = _gls_setup(decay_data, kind)
+        x_calc = data._gls_calc(zam, e, decay_data, kind="cumulative")
```
luca-fiorito-11, line 492: let's use some common nomenclature...and please adjust also `gls.py`.

MODEL
x : model input
y : model output
y=f(x) : model

PRIOR
x_prior : 1D prior data, e.g. IFY
Vx_prior : 2D covariance matrix of x_prior
S : 2D sensitivity of the model y=f(x)
y_calc : 1D calculated output using S.dot(x_prior), e.g. calculated CFY
Vy_calc : 2D calculated output using S.T.dot(Vx_prior.dot(S))

EXTRA
y_extra : 1D extra info on output
Vy_extra : 2D covariance matrix for y_extra


you often use `x` for the model output (like in this case), and then it gets messy

luca-fiorito-11 (changes requested):
I stopped the review.
Can you first change the nomenclature to a standard one?

AitorBengoechea:
I think all the nomenclature is changed according to the comment. The only part that i have not change is fy.py line 484, because i thought that in this way is more visual

AitorBengoechea:
I find a bug, i will upload latter the version without the bug

AitorBengoechea:
I upload everything, have a look when you can @luca-fiorito-11

sandy/core/cov.py
```diff
@@ -744,19 +744,19 @@ def gls_update(self,  S, Vy, threshold=None):
         1 -3.14286e-01  1.71429e-01
         """
         index, columns = self.data.index, self.data.columns
-        Vx = self.data.values
-        A = self._gls_cov_sensitivity(S, Vy, threshold=threshold).values
-        V_new = Vx - A.dot(Vx)
+        Vx_prior = self.data.values
+        A = self._gls_cov_sensitivity(S, Vy_extra, threshold=threshold).values
+        V_new = Vx_prior - A.dot(Vx_prior)
```
luca-fiorito-11, line 749: `V_new` should be `Vx_post`

sandy/gls.py
```diff
@@ -11,31 +11,32 @@
         ]
 
 
-def gls_update(y, S, Vx, Vy, x, x_p, threshold=None):
+def gls_update(y, S, Vx_calc, Vy_extra, y_calc, y_extra, threshold=None):
```
luca-fiorito-11, line 14: y_calc should not be there...it should be calculated from S*x_prior

sandy/gls.py
```diff
@@ -11,31 +11,32 @@
...
     """
     Perform GlS update for a given variances, vectors and sensitivity.
 
     Parameters
     ----------
     y: 1D iterable
```
luca-fiorito-11, line 20: I guess this is wrong....it should be `x_prior`

luca-fiorito-11: change the nomenclature

sandy/gls.py
```diff
@@ -50,10 +51,13 @@ def gls_update(y, S, Vx, Vy, x, x_p, threshold=None):
...
     dtype: float64
     """
-    y_, index = pd.Series(y).values, pd.Series(y).index
-    x_, x_p_ = pd.Series(x).values, pd.Series(x_p).values
-    delta = x_p_ - x_
-    Vx_ = sandy.CategoryCov(Vx)
-    A = Vx_._gls_general_sensitivity(S, Vy, threshold).values
+    y_ = pd.Series(y).values
+    index = pd.Series(y).index
+    y_calc_ = pd.Series(y_calc).values
```
luca-fiorito-11, line 56: `y_calc = S.dot(x_prior)`

sandy/gls.py
```diff
@@ -11,31 +11,32 @@
...
     Parameters
     ----------
     y: 1D iterable
         Vector in which we are going to apply GLS (MX1)
-    Vx : 2D iterable
-        Covariance vector (MXM).
-    Vy : 2D iterable
-        Extra Covariance vector (MXM).
+    Vx_calc : 2D iterable
+        2D covariance matrix of x_prior (MXN).
```
luca-fiorito-11, line 23: `Vx_prior`, not `Vx_calc`

sandy/gls.py
```diff
@@ -50,10 +51,13 @@ def gls_update(y, S, Vx, Vy, x, x_p, threshold=None):
...
-    A = Vx_._gls_general_sensitivity(S, Vy, threshold).values
+    y_ = pd.Series(y).values
+    index = pd.Series(y).index
+    y_calc_ = pd.Series(y_calc).values
+    y_extra_ = pd.Series(y_extra).values
+    delta = y_extra_ - y_calc_
+    Vx_calc_ = sandy.CategoryCov(Vx_calc)
+    S_ = pd.DataFrame(S).loc[index, pd.Series(y_calc).index]
+    A = Vx_calc_._gls_general_sensitivity(S_, Vy_extra, threshold).values
     y_new = y_ + A.dot(delta)
```
luca-fiorito-11, line 62: `x_post`, not `y_new`

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        TypeError
+            The kind is not implemented in the method.
+
+        Returns
+        -------
+        Vy_calc : `pd.DataFrame`
+            2D calculated output using S.T.dot(Vx_prior.dot(S)).
+
+        """
+        fy_data = self.filter_by('ZAM', zam).filter_by("E", e).data.set_index('ZAP')
```
luca-fiorito-11, line 470: you repeat this specific filter many times...maybe it can be a private method

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        if kind == 'cumulative':
+            Vx_prior = fy_data.query('MT==454').DFY
+            Vx_prior = sandy.CategoryCov.from_var(Vx_prior)
+            S = decay_data.get_qmatrix().T
```
luca-fiorito-11, line 474: @AitorBengoechea, I wonder if `decay_data.get_qmatrix()` should already return the transpose...because in theory you always need to transpose it if you want to use it...or not?

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            Vy_calc = Vx_prior.sandwich(S, threshold=threshold)
+        else:
+            raise TypeError('The kind introduced is not valid')
```
luca-fiorito-11, line 477: `'Keyword argument "kind" is not valid'`

luca-fiorito-11: it should be a `ValueError`

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        Vy_calc = self._gls_Vy_calc(zam, e, decay_data, kind, threshold=threshold)
+        # Create the Cov object:
+        Vy_calc = sandy.CategoryCov(Vy_calc)
+        # Fix the S with the correct dimension:
+        S = _gls_setup(decay_data, kind).loc[Vy_calc.data.index, Vy_calc.data.columns]
+        return Vy_calc.gls_update(S, Vy_extra, threshold=threshold).data
+
+    def _gls_y_calc(self, zam, e, decay_data, kind):
+        """
+        “a priori” calculated value for the vector of GLS.
```
luca-fiorito-11, line 409: "a priori" calculated model observable to adopt in GLS update.

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        Returns
+        -------
+        y_calc : `pandas.Series`
+            1D calculated output using S.dot(x_prior), e.g. calculated CFY
+
+        """
+        fy_data = self.filter_by('ZAM', zam).filter_by("E", e)
+        if kind == 'cumulative':
+            y_calc = fy_data.apply_qmatrix(zam, e, decay_data).data
+            # Apropiate format
```
luca-fiorito-11, line 437: appropriate

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            Keyword for obtaining sensitivity. The
+            default is 'cumulative'.
+        threshold : `int`, optional
+            Optional argument to avoid numerical fluctuations or
+            values so small that they do not have to be taken into
+            account. The default is None.
+
+        Returns
+        -------
+        `FY`
```
luca-fiorito-11, line 508: `sandy.Fy`

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+            IFY updated with GLS for a given zam, energy, decay_data and
+            new information.
+
+        """
+        data = self
```
luca-fiorito-11, line 513: i don't like much this....

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        # Filter FY data:
+        fy_data = data.filter_by('ZAM', zam)\
+                      .filter_by("E", e).data\
+                      .set_index('ZAP')[['MT', 'FY', 'DFY']]
+        # Divide the data:
+        IFY = fy_data.query('MT==454').FY
```
luca-fiorito-11, line 519: call it `x_prior`

sandy/fy.py
```diff
@@ -350,13 +352,182 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        mask = (data.data.ZAM == zam) & (data.data.MT == 454) & (data.data.E == e)
+        # Find the GLS varibles:
+        S = _gls_setup(decay_data, kind)
+        y_calc = data._gls_y_calc(zam, e, decay_data, kind)
+        Vy_calc = data._gls_Vy_calc(zam, e, decay_data, kind, threshold=threshold)
+        # Perform GLS:
+        IFY_new = sandy.gls_update(IFY, S, Vy_calc, Vy_extra, y_calc, y_extra,
```
luca-fiorito-11, line 526: `x_post`

sandy/fy.py
```diff
@@ -524,6 +695,35 @@ def to_hdf5(self, file, library):
 #        pass
 
 
+def _gls_setup(decay_data, kind, index=None, columns=None):
```
luca-fiorito-11, line 698: remove index and columns

sandy/fy.py
```diff
@@ -524,6 +695,35 @@ def to_hdf5(self, file, library):
...
+    Returns
+    -------
+    S : `pandas.DataFrame`
+        The sensitivity for performing GLS.
+
+    """
+    if kind == 'cumulative':
+        S = decay_data.get_qmatrix().T
+    else:
+        raise TypeError('The kind introduced is not valid')
```
luca-fiorito-11, line 723: same message reported in previous comment

sandy/fy.py
```diff
@@ -524,6 +695,35 @@ def to_hdf5(self, file, library):
...
+        The kind is not implemented in the method.
+
+    Returns
+    -------
+    S : `pandas.DataFrame`
+        The sensitivity for performing GLS.
+
+    """
+    if kind == 'cumulative':
+        S = decay_data.get_qmatrix().T
```
luca-fiorito-11, line 721: same comment as above

luca-fiorito-11 (changes requested):

sandy/fy.py
AitorBengoechea: I change the GLS method to be the same of the book Probability, stadistics, and Data Uncertainties in Nuclear Science and technology.

sandy/core/cov.py
```diff
@@ -621,6 +621,35 @@ def from_stdev(cls, std):
         var = std_ * std_
         return CategoryCov.from_var(var)
 
+    def _gls_Vy_calc(self, S):
+        """
+        2D calculated output using S.T.dot(Vx_prior).dot(S)
```
luca-fiorito-11, line 626: Here the transpose is not in the same place than in the method

sandy/core/cov.py
```diff
@@ -621,6 +621,35 @@ def from_stdev(cls, std):
...
         return CategoryCov.from_var(var)
 
+    def _gls_Vy_calc(self, S):
+        """
+        2D calculated output using S.T.dot(Vx_prior).dot(S)
+
+        Parameters
+        ----------
+        S : 2D iterable
+            Sensitivity square matrix (MXM).
```
luca-fiorito-11, line 631: Not square

sandy/gls.py
```diff
@@ -11,15 +11,15 @@
         ]
 
 
-def gls_update(y, S, Vx_calc, Vy_extra, y_calc, y_extra, threshold=None):
+def gls_update(x_prior, S, Vx_prior, Vy_extra, y_calc, y_extra, threshold=None):
```
luca-fiorito-11, line 14: `y_calc` shouldn't be an input

sandy/gls.py
```diff
@@ -51,13 +51,39 @@ def gls_update(y, S, Vx_calc, Vy_extra, y_calc, y_extra, threshold=None):
...
+        2D sensitivity of the model y=f(x) (MXN).
+
+    Returns
+    -------
+    y_calc : `pd.Series`
+        1D calculated output using S.dot(x_prior), e.g. calculated CFY
+
+    """
+    S_ = pd.DataFrame(S)
+    x_prior_ = pd.Series(x_prior).reindex(S_.columns).fillna(0)
```
luca-fiorito-11, line 87: I am not sure I like the reindex here...I would have expected the user to give the right dimensions

sandy/gls.py
```diff
@@ -51,13 +51,39 @@ def gls_update(y, S, Vx_calc, Vy_extra, y_calc, y_extra, threshold=None):
...
-    y_ = pd.Series(y).values
-    index = pd.Series(y).index
-    y_calc_ = pd.Series(y_calc).values
-    y_extra_ = pd.Series(y_extra).values
+    # Model calculus:
+    y_calc_ = _y_calc(x_prior, S)
+    # Data in a appropiate format
+    x_prior_ = pd.Series(x_prior)
+    y_extra_ = pd.Series(y_extra)
+    y_calc_ = y_calc_.reindex(y_extra_.index)
```
luca-fiorito-11, line 59: I am not sue about the reindex...the user should provide correct inidces

sandy/gls.py
```diff
@@ -51,13 +51,39 @@ def gls_update(y, S, Vx_calc, Vy_extra, y_calc, y_extra, threshold=None):
...
     delta = y_extra_ - y_calc_
-    Vx_calc_ = sandy.CategoryCov(Vx_calc)
-    S_ = pd.DataFrame(S).loc[index, pd.Series(y_calc).index]
-    A = Vx_calc_._gls_general_sensitivity(S_, Vy_extra, threshold).values
-    y_new = y_ + A.dot(delta)
-    return pd.Series(y_new, index=index)
+    Vx_prior_ = sandy.CategoryCov(Vx_prior)
+    S_ = pd.DataFrame(S).loc[Vx_prior_.data.index, Vx_prior_.data.columns]
```
luca-fiorito-11, line 62: Same here, dimensions should be correct

luca-fiorito-11 (changes requested):
The only doubt is about `loc` and `reindex`. @AitorBengoechea  Do you think we should assume that the user always gives consistent dimensions? If we decide of reindexing, then more tests are needed.

sandy/core/cov.py
AitorBengoechea: I dont know what is the difference. In the probability statistics and data uncertainties in nuclear science and technology appears in this way, and the model is y_a=f(theta_a). In our case is y_a=Q*theta_a, so i copy because i dont really the difference

luca-fiorito-11: I mean that in the docstring and in the function you don't do the same...probably you just need to fix the docstring

AitorBengoechea:
The problem is: 
In the model, the calculus of CFY gives 3850 results. So it not cosistent with the nucluar library. So I introduce a argument for reindexing. 
With this argument, fy.py only has to give S.columns==x_prior.index. Then, if this argument is active, after obtaining y_calc, begins the reindexing for y_calc(now 983) and for S(now, 983,983). In this way, only activating this argument, we obtain a consintent model to CFY = Q*IFY without loosing globality. Now, all fix with each other and Category.Cov works.

luca-fiorito-11:
> The problem is: In the model, the calculus of CFY gives 3850 results. So it not cosistent with the nucluar library. So I introduce a argument for reindexing. With this argument, fy.py only has to give S.columns==x_prior.index. Then, if this argument is active, after obtaining y_calc, begins the reindexing for y_calc(now 983) and for S(now, 983,983). In this way, only activating this argument, we obtain a consintent model to CFY = Q*IFY without loosing globality. Now, all fix with each other and Category.Cov works.

I agree with you

luca-fiorito-11:
> > The problem is: In the model, the calculus of CFY gives 3850 results. So it not cosistent with the nucluar library. So I introduce a argument for reindexing. With this argument, fy.py only has to give S.columns==x_prior.index. Then, if this argument is active, after obtaining y_calc, begins the reindexing for y_calc(now 983) and for S(now, 983,983). In this way, only activating this argument, we obtain a consintent model to CFY = Q*IFY without loosing globality. Now, all fix with each other and Category.Cov works.
> 
> I agree with you

But then, maybe we should just do it for the FY object and not for `gls.py`, which in my opinion should be a more general function...what do you think?

Either way, we can always change it later...just add some simple tests where loc and reindex are used

sandy/gls.py
AitorBengoechea: Reason for putting there the reindex:
In the problem we are solving, the dimensions are for qmatrix(3850X983, after passing through the loc that makes multiplication possible) and IFY(983X1). Therefore, CFY will have 3850X1. 
According to my tests, the 983 values that appear in CFY reducing the vector 3850X1 are closer to the values of the library than the ones we get if we do qmatrix(983X983)*IFY(983X1). 
For this reason, here I expand IFY.

AitorBengoechea:
Assumptions:
The user is entering as data S, x_prior and y_extra. we assume that he has in mind y(MX1) = S(MXN)*x(NX1).
Conclusions:
We already know in advance the dimensions of y_calc, given that y_calc(MX1) = y_extra(MX1). Therefore, we can remove the (if reindex is true). With (y_calc.reindex) and (S_loc) we make sure that the model is being fulfilled, according to the data we are entering (y_extra). Therefore, I would not make any assumptions, but rely on the data that the user enters. 
On the other hand, following this logic, if we put a S.loc[:, x_prior.index] in gls_update, we make sure that S(MXN)*x(NX1) is fulfilled based only on the data provided.
What do you think about this?

sandy/fy.py
```diff
@@ -213,6 +213,54 @@ def _expand_zam(self):
                            dtype=int)
         return self.data.assign(Z=zam.Z, A=zam.A, M=zam.M)
 
+    def get_decay_chains_sensitivity(self, zam, e):
```
luca-fiorito-11, line 216: Nice!

sandy/fy.py
```diff
@@ -296,13 +344,15 @@ def apply_bmatrix(self, zam, e, decay_data):
             .filter_by('MT', 459)\
             .filter_by("E", e).data\
             .set_index('ZAP')['FY']
```
luca-fiorito-11, line 346: call `_filters`

sandy/fy.py
```diff
@@ -350,13 +400,171 @@ def apply_qmatrix(self, zam, energy, decay_data):
             .filter_by('MT', 454)\
             .filter_by("E", energy).data\
             .set_index('ZAP')['FY']
```
luca-fiorito-11, line 402: call `_filters`

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -789,9 +850,26 @@ def sandwich(self, S):
...
+        2	0.00000e+00	0.00000e+00	2.70000e+01
         """
-        C = self.data
-        return corr2cov(C, S)
+        if pd.DataFrame(S).shape[1] == 1:
+            C = self.data
+            sandwich = corr2cov(C, S)
+        else:
+            S_ = pd.DataFrame(S).T
+            sandwich = self._gls_Vy_calc(S_)
```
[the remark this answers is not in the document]
AitorBengoechea: 1D sensitivity goes to corr2cov(C, S). 2D goes to self._gls_Vy_calc(S_)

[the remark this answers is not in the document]
AitorBengoechea: Because pd.DataFrame(S).shape[1] measure the number of columns of pd.DataFrame(S).values

sandy/gls.py
```diff
@@ -3,39 +3,47 @@
...
-        Sensitivity square matrix (MXM).
-    x : 1D iterable
-        x vector.
-    x_p : 1D iterable
-        x vector perturbed.
+        2D sensitivity of the model y=f(x) (MXN).
+    y_extra : 1D iterable
+        1D extra info on output (NX1)
     threshold : `int`, optional
             Thereshold to avoid numerical fluctuations. The default is None.
```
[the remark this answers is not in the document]
AitorBengoechea: Is not need. We belive in the user and fix S and y_calc according to the information given by the user:
y_calc : same dimensions that y_extra
S : y_calc = S * x_prior, so the columns are fixed with x_prior and the row are fixed with y_calc, that has the same dimensions that y_extra

sandy/core/cov.py
```diff
@@ -694,35 +755,35 @@ def _gls_cov_sensitivity(self, S, Vy, threshold=None):
...
         >>> Vy = pd.DataFrame([[1, 0], [0, 1]], index=[1, 2], columns=[1, 2])
         >>> var._gls_cov_sensitivity(S, Vy)
                       1	          2
-        3	2.57143e-01	3.14286e-01
-        4	3.14286e-01	8.28571e-01
+        1	4.00000e-01	4.00000e-01
+        2	4.00000e-01	6.85714e-01
         """
-        index, columns = pd.DataFrame(S).index, pd.DataFrame(S).columns
+        index, columns = pd.DataFrame(S).columns, pd.DataFrame(S).columns
```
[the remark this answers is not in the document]
AitorBengoechea: Yes, is correct. I did all the index and columns in all the step by hand.

sandy/gls.py
AitorBengoechea: The user should provide correct index. If dont pass correct index, the program will broke, avoiding the user to obtain results without sense in the best of the cases

sandy/gls.py
```diff
@@ -60,9 +60,10 @@ def gls_update(x_prior, S, Vx_prior, Vy_extra, y_extra,
...
     """
     # Model calculus:
+    x_prior_ = pd.Series(x_prior)
     S_ = pd.DataFrame(S)
+    S_ = S_.loc[:, x_prior.index]
     y_calc_ = _y_calc(x_prior, S)
-    x_prior_ = pd.Series(x_prior)
     y_extra_ = pd.Series(y_extra)
     y_calc_ = y_calc_.reindex(y_extra_.index)
     S_ = S_.loc[y_extra_.index, :]
```
luca-fiorito-11, line 69: why `S_ = S_.loc[:, x_prior.index]` and `S_ = S_.loc[y_extra_.index, :]` are on two different lines?
Could it be `S_ = S_.loc[y_extra_.index, x_prior.index]`?

AitorBengoechea: They are in different lines because i dont want to lose accuracy: In our specific problem, with S_ = S_.loc[y_extra_.index, x_prior.index] the y_calc are more different from the libraries that if we do in separate lines, because we are using more information of qmatrix to calculate y_calc

luca-fiorito-11 (approved):

PR #107, opened 2021-12-16
corrections

AitorBengoechea:
As you may see in the test, in minimal (mt=454) at the beginning the nucleus 621480 doesnt have CFY and after applyqmatrix has

### DIFF
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -442,7 +442,7 @@ class Fy():
         data = pd.concat([data, data.loc[mask]]).drop_duplicates(keep=False)
         fy_data = fy_data.reindex(Q.columns).fillna(0)
         # Apply qmatrix
-        cfy_calc_values = Q.dot(fy_data).rename('FY').to_frame().reset_index().rename(columns={'DAUGHTER': 'ZAP'})
+        cfy_calc_values = Q.dot(fy_data).rename('FY').reset_index().rename(columns={'DAUGHTER': 'ZAP'})
         # Calculus in appropiate way:
         cfy_calc_values[['MAT', 'ZAM', 'MT', 'E', 'DFY']] = [mat, zam, 459, energy, 0]
         data = pd.concat([data, cfy_calc_values], ignore_index=True)
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -422,23 +422,32 @@ class Fy():
         >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
         >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
         >>> npfy = Fy(minimal_fytest_2)
-        >>> npfy_pert = npfy.apply_bmatrix(942390, 5.00000e+05, decay_fytest)
-        >>> diff = npfy_pert.data[npfy_pert.data.FY != npfy.data.FY].FY
-        >>> comp = npfy_pert.data.query('ZAM==942390 & MT==459 & E==500e3').squeeze().FY
-        >>> assert comp.values.all() == diff.values.all()
-
+        >>> npfy_pert = npfy.apply_qmatrix(942390, 5.00000e+05, decay_fytest)
+        >>> npfy_pert.data[npfy_pert.data.MT == 459]
+             MAT	 MT	   ZAM	   ZAP	          E	         FY	        DFY
+        3	9437	459	942390	591480	5.00000e+05	1.00000e-01	0.00000e+00
+        4	9437	459	942390	591481	5.00000e+05	2.00000e-01	0.00000e+00
+        5	9437	459	942390	601480	5.00000e+05	6.00000e-01	0.00000e+00
+        6	9437	459	942390	621480	5.00000e+05	6.00000e-01	0.00000e+00
         """
-        new_data = self.data.copy()
+        # Obtain the data:
+        data = self.data.copy()
         conditions = {'ZAM': zam, 'MT': 454, "E": energy}
-        fy_data = self._filters(conditions).data.set_index('ZAP')['FY']
-        index = fy_data.index
+        fy_data = self._filters(conditions).data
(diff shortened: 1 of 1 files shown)


sandy/fy.py
```diff
@@ -422,23 +422,32 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        5	9437	459	942390	601480	5.00000e+05	6.00000e-01	0.00000e+00
+        6	9437	459	942390	621480	5.00000e+05	6.00000e-01	0.00000e+00
+        """
+        # Obtain the data:
+        data = self.data.copy()
         conditions = {'ZAM': zam, 'MT': 454, "E": energy}
-        fy_data = self._filters(conditions).data.set_index('ZAP')['FY']
-        index = fy_data.index
+        fy_data = self._filters(conditions).data
+        mat = fy_data.iloc[0, 0]
```
luca-fiorito-11, line 437: it is better `mat = fy_data.MAT.squeeze()`

sandy/fy.py
```diff
@@ -422,23 +422,32 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
+        fy_data = fy_data.set_index('ZAP')['FY']
         Q = decay_data.get_qmatrix()
+        # Put the data in a approppiate format:
+        mask = (data.ZAM == zam) & (data.MT == 459) & (data.E == energy)
+        data = pd.concat([data, data.loc[mask]]).drop_duplicates(keep=False)
```
luca-fiorito-11, line 442: why are you doing this?

sandy/fy.py
```diff
@@ -422,23 +422,32 @@ def apply_qmatrix(self, zam, energy, decay_data):
...
         fy_data = fy_data.reindex(Q.columns).fillna(0)
         # Apply qmatrix
-        fy_calc = Q.dot(fy_data).loc[index]
-        mask = (new_data.ZAM == zam) & (new_data.MT == 459) & (new_data.E == energy)
-        new_data.loc[mask, 'FY'] = fy_calc.values
-        return self.__class__(new_data)
+        fy_calc_values = Q.dot(fy_data).rename('FY').to_frame()
```
luca-fiorito-11, line 445: better call it `cfy_calc_values`

luca-fiorito-11: put lines 445 and 446 together:
`cfy_calc_values = Q.dot(fy_data).rename('FY').reset_index().rename(columns={'DAUGHTER': 'ZAP'}`

if you use `reset_index` you don't need to use `to_frame`

sandy/fy.py
AitorBengoechea: To remove the duplicate data

sandy/fy.py
AitorBengoechea: But reset_index only work in a dataFrame and you obtain a pd.Series from the multiplication

sandy/fy.py
AitorBengoechea: Dont work, i thought is because the have different shapes

sandy/fy.py
luca-fiorito-11: https://pandas.pydata.org/docs/reference/api/pandas.Series.reset_index.html

sandy/fy.py
luca-fiorito-11: what duplicate data? why do you need concat?

AitorBengoechea: To remove only data.loc[mask].

sandy/fy.py
luca-fiorito-11: well, at least use `fy_data.MAT.iloc[0]`

sandy/fy.py
luca-fiorito-11: I still don't get it. `data.loc[mask]` is already in `data`, then why do you duplicate it using `concat`? and why then you drop the duplicates?

luca-fiorito-11 (changes requested):

sandy/fy.py
AitorBengoechea: Because is the only way that i found in internet to drop from a dataFrame another dataFrame

luca-fiorito-11: but why do you want to do it? what do you want to do? if you completely delete this line the code would still work fine, right? or not?

AitorBengoechea: No, in this line i am droping all cumulative fission yields defined by the mask. Then, when i use concat, the enter because the data that is suppose to be replacing is not there

luca-fiorito-11: what is `pd.concat([data, data.loc[mask]])` doing? it concatenates duplicates of the CFY selected by the mask, right? those CFYs already existed in data

AitorBengoechea: The problem was that i was not able to erase the row containing the information of data.loc[mask]]. So i found in the Internet the following method. First fetch the two data, duplicating CFY selected by the mask. Then, with drop_duplicates the duplicated files will be erased. 
In this way, after this line, the data doesnt have the information of the CFY selected by the mask

luca-fiorito-11: Now I see what you are trying to do...
It should be something like `data = data.loc[~mask]`

sandy/fy.py
luca-fiorito-11: This is also still open...why not `reset_index` instead of `to_frame().reset_index`?
It doesn't work?

sandy/fy.py
AitorBengoechea: Yes !!!! I changed now, thanks

sandy/fy.py
luca-fiorito-11: @AitorBengoechea , this was not resolved? is it now?

AitorBengoechea: Yes, now it returns in the output the 3850 nucleus that are in the rows of the Q matrix.

PR #108, opened 2021-12-16
This is what i have for the moment

AitorBengoechea:


AitorBengoechea:
I put here, but for the moment is very simple and without docstrings. Docstring will be upload with the next methods

PR #110, opened 2021-12-18
Sparse

AitorBengoechea:


sandy/core/cov.py
```diff
@@ -1124,6 +1314,53 @@ def random_cov(cls, size, stdmin=0.0, stdmax=1.0, correlations=True,
...
+        elif method == 'coo_matrix':
+            data_sp = sps.coo_matrix(data)
+        elif method == 'csc_matrix':
+            data_sp = sps.csc_matrix(data)
+        elif method == 'dia_matrix':
+            data_sp = sps.dia_matrix(data)
+        elif method == 'dok_matrix':
+            data_sp = sps.dok_matrix(data)
+        elif method == 'lil_matrix':
+            data_sp = sps.lil_matrix(data)
```
luca-fiorito-11, line 1362: Should we add a else and raise a Value Error?

sandy/core/cov.py
```diff
@@ -1649,6 +1886,20 @@ def get_eig(cov, sort=True):
     return E, V
 
 
+def get_eig_sparse(cov, sort=True):
+    try:
+        E, V = sps.linalg.eig(cov)
+    except:
+        E, V = scipy.linalg.eig(cov.toarray())
```
luca-fiorito-11, line 1893: Why try/except? Should we add it also to get_eig?

sandy/core/cov.py
```diff
@@ -1124,6 +1314,53 @@ def random_cov(cls, size, stdmin=0.0, stdmax=1.0, correlations=True,
         std = np.random.uniform(stdmin, stdmax, size)
         return cls.corr2cov(corr, std, **kwargs)
 
+    def to_sparse(self, method='csr_matrix'):
+        """
+        Method to obtain sparse matrix
```
luca-fiorito-11, line 1319: Method to extract categorycov values into a sparse matrix

sandy/core/cov.py
```diff
@@ -470,14 +496,27 @@ def invert(self):
...
         """
         index, columns = self.data.index, self.data.columns
-        M_nonzero_idxs, M_reduce = self._reduce_size()
-        unit = np.identity(len(M_reduce))
-        M_reduce_inv = splu(csc_matrix(M_reduce)).solve(unit)
-        data = CategoryCov._restore_size(M_nonzero_idxs, M_reduce_inv,
-                                         len(self.data))
-        M_inv = pd.DataFrame(data.data,
+        if sparse is True:
+            cov = self.to_sparse(method='coo_matrix').tocsc()
```
luca-fiorito-11, line 509: Why not `self.to_sparse(method='csc_matrix')`?

sandy/core/cov.py
```diff
@@ -470,14 +496,27 @@ def invert(self):
...
+        2 0.00000e+00 0.00000e+00 3.30000e-01
         """
         index, columns = self.data.index, self.data.columns
-        M_nonzero_idxs, M_reduce = self._reduce_size()
-        unit = np.identity(len(M_reduce))
-        M_reduce_inv = splu(csc_matrix(M_reduce)).solve(unit)
-        data = CategoryCov._restore_size(M_nonzero_idxs, M_reduce_inv,
-                                         len(self.data))
-        M_inv = pd.DataFrame(data.data,
+        if sparse is True:
```
luca-fiorito-11, line 508: `if sparse:`

sandy/core/cov.py
```diff
@@ -470,14 +496,27 @@ def invert(self):
...
+            cov = self.to_sparse(method='coo_matrix').tocsc()
+            lu = spsl.splu(cov)
+            eye = np.eye(cov.shape[0])
+            data = lu.solve(eye)
```
luca-fiorito-11, line 512: You don't need to reduce and restore?

sandy/core/cov.py
```diff
@@ -470,14 +496,27 @@ def invert(self):
...
+        else:
+            M_nonzero_idxs, M_reduce = self._reduce_size()
+            unit = np.identity(len(M_reduce))
+            M_reduce_inv = splu(csc_matrix(M_reduce)).solve(unit)
```
luca-fiorito-11, line 516: This is also using a sparse matrix, although we are in `else`

sandy/core/cov.py
```diff
@@ -531,24 +575,33 @@ def decompose(self):
...
+        if sparse==True:
+            M = sps.csr_matrix(V).dot(sps.diags(np.sqrt(E))).toarray()
+        else:
+            M = V.values.dot(np.diag(np.sqrt(E.values)))
         Q, R = scipy.linalg.qr(M.T)
         return R.T
 
     @classmethod
-    def from_var(cls, var):
+    def from_var(cls, var, sparse=False):
```
luca-fiorito-11, line 595: Why do we need sparse here?

sandy/core/cov.py
```diff
@@ -576,21 +629,34 @@ def from_var(cls, var):
...
+            cov_values = sps.diags(var_.values).toarray()
+        else:
+            cov_values = np.diag(var_)
+        cov = pd.DataFrame(cov_values,
                            index=var_.index, columns=var_.index)
         return cls(cov)
 
     @classmethod
-    def from_stdev(cls, std):
+    def from_stdev(cls, std, sparse=False):
```
luca-fiorito-11, line 650: Why do we need sparse here?

sandy/core/cov.py
```diff
@@ -687,15 +776,29 @@ def _gls_G(self, S, Vy_extra):
...
+        >>> Vy = np.diag(pd.Series([1, 1]))
+        >>> sensitivity._gls_G(S, Vy, sparse=True)
+                    	0	      1
+        0	6.00000e+00	1.10000e+01
+        1	1.10000e+01	2.60000e+01
         """
         index = pd.DataFrame(Vy_extra).index  # Simetry of cov matrix
         Vy_extra_ = sandy.CategoryCov(Vy_extra).data.values
         # GLS_sensitivity:
         Vy_calc = self._gls_Vy_calc(S).values
```
luca-fiorito-11, line 791: Sparse should be passed here

sandy/core/cov.py
```diff
@@ -687,15 +776,29 @@ def _gls_G(self, S, Vy_extra):
...
+        if sparse is True:
+            Vy_calc = sps.csr_matrix(Vy_calc)
```
luca-fiorito-11, line 793: Thos is not really necessary, right?!

sandy/core/cov.py
```diff
@@ -725,13 +830,21 @@ def _gls_G_inv(self, S, Vy_extra):
...
+        >>> S = np.array([[1, 2], [3, 4]])
+        >>> sensitivity = sandy.CategoryCov.from_var([1, 1])
+        >>> Vy = np.diag(pd.Series([1, 1]))
+        >>> sensitivity._gls_G_inv(S, Vy, sparse=True)
+                      0	              1
+        0	7.42857e-01	    -3.14286e-01
+        1	-3.14286e-01	1.71429e-01
         """
         index = pd.DataFrame(Vy_extra).index
         G = self._gls_G(S, Vy_extra).values
```
luca-fiorito-11, line 843: You should pass sparse here, right?!

sandy/core/cov.py
```diff
@@ -812,22 +941,35 @@ def _constrained_gls_sensitivity(self, S, threshold=None):
...
         S_ = pd.DataFrame(S)
         index = S_.index
         columns = S_.columns
-        # constrained Least Squares sensitivity
-        G_inv = self._gls_G_inv(S, Vy_extra).values
-        sensitivity = G_inv.dot(S_.values).dot(Vx_prior)
+        if sparse is True:
+            Vx_prior = self.to_sparse()
+            S_ = sps.csr_matrix(S_.values)
+            G = self._gls_Vy_calc(S, sparse=sparse)
```
luca-fiorito-11, line 956: Why don't you call self._gls_G_inv?

sandy/core/cov.py
```diff
@@ -812,22 +941,35 @@ def _constrained_gls_sensitivity(self, S, threshold=None):
...
+            G_inv = sandy.CategoryCov(G).invert(sparse=sparse)
+            G_inv =  sps.csr_matrix(G_inv.data.values)
+            # constrained Least Squares sensitivity
+            sensitivity = G_inv.dot(S_).dot(Vx_prior).toarray()
+        else:
+            Vx_prior = self.data.values
+            Vy_extra = pd.DataFrame(0, index=self.data.index,
+                                    columns=self.data.columns)
+            # constrained Least Squares sensitivity
```
luca-fiorito-11, line 965: Instead of passing a df with zeros, should Vy_extra be optional in _gls_G_inv?

sandy/core/cov.py
```diff
@@ -906,14 +1063,27 @@ def gls_update(self, S, Vy_extra, threshold=None):
...
         1 -4.00000e-01  3.14286e-01
+        >>> var.gls_update(S, Vy, sparse=True)
+                     0            1
+        0  6.00000e-01 -4.00000e-01
+        1 -4.00000e-01  3.14286e-01
         """
         index, columns = self.data.index, self.data.columns
-        Vx_prior = self.data.values
-        A = self._gls_cov_sensitivity(S, Vy_extra, threshold=threshold).values
+        if sparse is True:
```
luca-fiorito-11, line 1072: I would just do 1 if/else, not 2

sandy/core/cov.py
```diff
@@ -945,12 +1117,26 @@ def constrained_gls_update(self, S, threshold=None):
...
+            general_sens = self._constrained_gls_sensitivity(S, sparse=sparse, 
+                                                             threshold=threshold)
+            general_sens = sps.csr_matrix(general_sens)
+            A = Vx_prior.dot(S_.T).dot(general_sens)
+            Vx_post = Vx_prior - A.dot(Vx_prior)
+            Vx_post = Vx_post.toarray()
+        else:
+            Vy_extra = pd.DataFrame(0, index=self.data.index,
+                                    columns=self.data.columns)
+            Vx_post = self.gls_update(S, Vy_extra, threshold=threshold).data.values
```
luca-fiorito-11, line 1136: Same as above, maybe Vy_extra should be optional in gls_update

luca-fiorito-11 (changes requested):
Good job!

sandy/core/cov.py
AitorBengoechea: Because depending on the matrix cov, sometimes scipy.sparse is not optimize for that matrix and raise an error saying to move to scipy.linalg.eig.

sandy/core/cov.py
AitorBengoechea: Because i read that solving the inversion of a matrix in this way is more efficient than sps.invert(). But for use of this spsl.splu(cov), cov must be a sparse matrix in COOrdinate format

sandy/core/cov.py
AitorBengoechea: No, because the method is optimized for the inversion. Even if we use spsl.inv(), there is not need to reduce and restore, becuase sparse matrix only contain the non null values.

sandy/core/cov.py
AitorBengoechea: I thought that with sparse matrix is not need to reduce and restore size.

sandy/core/cov.py
AitorBengoechea: I tried to optimize with sparse matrix all the mathematical operations that we have. Creating a diagonal matrix I thought it consume less memory in that way.

sandy/core/cov.py
AitorBengoechea: Same as above

sandy/core/cov.py
AitorBengoechea: I think that all mathematical operation are more optimized for sparse matrices, because here only sum non null element. I numpy i thouht that sum all the elements, so we are avoiding a lot of operations

sandy/core/cov.py
luca-fiorito-11: So, why do you use `tocsc`? What type is your cov after `tocsc`?

sandy/core/cov.py
AitorBengoechea: Becuase is impossible. You cant pass as argument a matrix full of zeros in sparse form.

sandy/core/cov.py
luca-fiorito-11: Ok, can ou make sure of this with a test?

sandy/core/cov.py
AitorBengoechea: Better idea. So i will correct the above comment also

sandy/core/cov.py
luca-fiorito-11: I mean, we call csc_matrix that is sparse, although this is the solution for the non-sparse matrix.
Should we just do `splu(M_reduce).solve(unit)`?

sandy/core/cov.py
luca-fiorito-11: What exception do you get? You could work around a specific exception,  say `TypeError`, using `except TypeError`.

luca-fiorito-11: To avoid rewriting parts of code, can we use:
```
except Exception:
    return self.eig()
```

sandy/core/cov.py
luca-fiorito-11: Ok, but then, can you use only 1 if/else clause instead of 2? To be consistent everywhere in the module (sometimes you use 1 and other times 2)

sandy/core/cov.py
AitorBengoechea: Convert this matrix to Compressed Sparse Column format

AitorBengoechea: Now is better

sandy/core/cov.py
AitorBengoechea: Yes

AitorBengoechea:
For the moment i update also GLS method. I do not know if gls.py will need more optimization. When i calculate with my real test i will optimize more if i need. For the moment i don not correct fy.py apply_qmatrix and b matrix because it will go in another branch

sandy/core/cov.py
luca-fiorito-11: @AitorBengoechea , i still don't get why calling csc_matrix

sandy/core/cov.py
```diff
@@ -1360,6 +1366,8 @@ def to_sparse(self, method='csr_matrix'):
             data_sp = sps.dok_matrix(data)
         elif method == 'lil_matrix':
             data_sp = sps.lil_matrix(data)
+        else:
+            raise ValueError('The method not exist in scipy.sparse')
```
luca-fiorito-11, line 1370: The method does not exist in scipy.sparse

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: For the moment i will leave this part as it is, because sps.linalg.eig doesnt exits, so it was always doing scipy.linalg.eig(cov.toarray())

sandy/core/cov.py
AitorBengoechea: https://stackoverflow.com/questions/21693613/is-it-possible-to-compute-an-inverse-of-sparse-matrix-in-python-as-fast-as-in-ma 
I found here in the second answer. If you put csr, you will recibe the following SparseEfficiencyWarning: splu requires CSC matrix format

sandy/core/cov.py
```diff
@@ -764,7 +746,7 @@ def _gls_G(self, S, Vy_extra, sparse=False):
         ----------
         S : 2D iterable
             Sensitivity matrix (MXN).
-        Vy_extra : 2D iterable
+        Vy_extra : 2D iterable, optional.
```
luca-fiorito-11, line 749: optional is `None`

luca-fiorito-11: Missing test for Vy_extra = None

sandy/core/cov.py
```diff
@@ -278,16 +278,14 @@ def data(self, data):
     def size(self):
         return self.data.values.shape[0]
 
-    def eig(self, sort=True, sparse=False):
+    def eig(self, sort=True):
         """
```
luca-fiorito-11, line 282: Why is the sparse part gone?

sandy/core/cov.py
```diff
@@ -1894,20 +1877,6 @@ def get_eig(cov, sort=True):
     return E, V
 
 
-def get_eig_sparse(cov, sort=True):
```
luca-fiorito-11: Why is the sparse part gone?

sandy/core/cov.py
```diff
@@ -794,16 +776,19 @@ def _gls_G(self, S, Vy_extra, sparse=False):
         1	1.10000e+01	2.60000e+01
         """
         index = pd.DataFrame(Vy_extra).index  # Simetry of cov matrix
```
luca-fiorito-11, line 778: Symmetry

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: I change the entire format

sandy/core/cov.py
AitorBengoechea: Same as above

AitorBengoechea:
The new format is:
get_eig ==> Only one, because i was repeating the code. Now recive as optional argument sparse, to go to try/except. Inside this is one filters.warnings, becuase if you try to spsl.eigs(cov) that is not allow, firts print a runtime warning and the raise a TypeError o ValueError. So, i made a loop for that.
Firts test with sparse=True: I know that the error is going to raise, so is the except path.
Second test: eigs works.

sandy/core/cov.py
```diff
@@ -879,7 +924,7 @@ def _gls_general_sensitivity(self, S, Vy_extra=None, sparse=False,
         >>> S = pd.DataFrame([[1, 2], [3, 4]], columns=[1, 2],index=[3, 4])
         >>> sensitivity = sandy.CategoryCov.from_var([1, 1])
         >>> Vy = pd.DataFrame([[1, 0], [0, 1]], index=[1, 2], columns=[1, 2])
-        >>> sensitivity._gls_general_sensitivity(S, Vy)
+        >>> sensitivity._gls_general_sensitivity(S, Vy_extra=Vy)
```
luca-fiorito-11, line 927: We are also missing a test where Vy_extra is not given

sandy/core/cov.py
```diff
@@ -891,7 +936,7 @@ def _gls_general_sensitivity(self, S, Vy_extra=None, sparse=False,
         index, columns = pd.DataFrame(S).columns, pd.DataFrame(S).index
         S_ = pd.DataFrame(S).values
         # GLS_sensitivity:
-        G_inv = self._gls_G_inv(S, Vy_extra=Vy_extra, sparse=sparse).values
+        G_inv = self._gls_G_inv(S, Vy_extra, sparse=sparse).values
```
luca-fiorito-11, line 939: I preferred the previous line `G_inv = self._gls_G_inv(S, Vy_extra=Vy_extra, sparse=sparse).values`

luca-fiorito-11 (changes requested):

luca-fiorito-11:
@AitorBengoechea , why did you close this pull request? 
The new one doesn't keep track of all the changes and I have to go again through everything and rewrite all the comments. 
Can we finalize this first?

sandy/core/cov.py
```diff
@@ -478,7 +478,7 @@ def _restore_size(cls, nonzero_idxs, cov_reduced, dim):
             cov[ni, nonzero_idxs] = cov_reduced[i]
         return cls(cov)
 
-    def invert(self, sparse=False):
+    def invert(self):
```
luca-fiorito-11, line 481: Then parameter sparse should not appear anymore in the docstring

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -1638,8 +1907,15 @@ def random_ctg_cov(index, stdmin=0.0, stdmax=1.0, correlations=True, seed=None):
     return pd.DataFrame(cov, index=index, columns=index)
 
 
-def get_eig(cov, sort=True):
-    E, V = scipy.linalg.eig(cov)
+def get_eig(cov, sparse=False, sort=True):
```
luca-fiorito-11, line 1910: Last comment,  I think there's no need anymore to have a function `get_eig`.
This algorithm should be merged in method `eig` of `CategoryCov` (maybe being called `get_eigs`).

What do you think?

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: In my opinion is better to merge the algorithm in `eig` instead of create a new method.

luca-fiorito-11: Yes, in `eig`...I just don't like the name...I think calling it `get_eigs` is better

luca-fiorito-11 (approved):

PR #117, opened 2022-01-13
Pytables and sparse and some changes

AitorBengoechea:


AitorBengoechea:
@luca-fiorito-11 is ready for corrections

AitorBengoechea:
The last change is due to the fact that the lack of inconsistency between decay and fy is deeper than I first thought.

AitorBengoechea:
Cholesky descomposition to obtain low triangular matrix is now available in sparse way and sparse + pytables way

PR #118, opened 2022-01-23
Extract BR

EnricaBelfiore:


sandy/decay.py
```diff
@@ -132,6 +132,66 @@ def get_halflives(self):
...
+                RTYP.append(float(rtyp) / 10)
+                ZAM.append(zam)
+        tuples = list(zip(* [ZAM,
+                             RTYP]))
+        idx = pd.MultiIndex.from_tuples(tuples, names=['ZAM', 'RTYP'])
+        df = pd.DataFrame(br, index=idx, columns=['BR', 'DBR'])
+        if with_uncertainty:
+            return df
+        else:
+            return df.drop(['DBR'], axis=1)
```
luca-fiorito-11, line 193: For consistency with `get_decayconstant` I would do
```
return df.BR
```
which is a series.
But then the Returns in the docstring should be changed.

sandy/decay.py
```diff
@@ -132,6 +132,66 @@ def get_halflives(self):
...
+        for zam, dic in self.data.items():
+            for rtyp, dk in dic['decay_modes']:
+                br.append([
+                    dk['branching_ratio'],
+                    dk['branching_ratio_uncertainty'],
+                    ])
+                RTYP.append(float(rtyp) / 10)
+                ZAM.append(zam)
+        tuples = list(zip(* [ZAM,
+                             RTYP]))
```
luca-fiorito-11, line 187: Not sure the list is needed

sandy/decay.py
```diff
@@ -132,6 +132,66 @@ def get_halflives(self):
...
+        ZAM = []
+        RTYP = []
+        for zam, dic in self.data.items():
+            for rtyp, dk in dic['decay_modes']:
+                br.append([
+                    dk['branching_ratio'],
+                    dk['branching_ratio_uncertainty'],
+                    ])
+                RTYP.append(float(rtyp) / 10)
+                ZAM.append(zam)
```
luca-fiorito-11, line 185: Why is rtyp a float? Can it be a int? Why do you divide by 10?

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: It is usually a float for multiple particle decay and I divided by 10 to get the rtyp value stored in ENDF6. To solve it I changed the method read_rdd

luca-fiorito-11: read_rdd and mf8 must not be changed this way.
Please, restore them as they originally were.

Check again what rtyp is on the endf6 manual.
As we have already discussed, if you divide by 10 you lose info on the gamma decay...it also won't work for delayed neutron precursors

luca-fiorito-11 (changes requested):

sandy/sections/mf8.py
```diff
@@ -544,8 +546,8 @@ def _write_rdd(sec):
     65146100.0 44179.7000          0          0          6          01803 8457    2
     163850.000 403.088000 1554660.00 895.993000 0.00000000 0.000000001803 8457    3
     4.00000000 1.00000000          0          0         12          21803 8457    4
-    2.00000000 0.00000000 1233300.00 800.000000 3.000000-6 1.000000-61803 8457    5
-    1.00000000 0.00000000 2058980.00 330.000000 9.999970-1 1.000000-61803 8457    6
+    20.0000000 0.00000000 1233300.00 800.000000 3.000000-6 1.000000-61803 8457    5
```
luca-fiorito-11, line 549: the first number should be `2.00000000`, not `20.0000000`

luca-fiorito-11 (changes requested):

sandy/sections/mf8.py
EnricaBelfiore: More or less... do you mean you prefer to change the rtype format here instead of in to_endf6 method? In the second example of this method there are two rtyps that should be 4.0 and 6.0 instead of 40.0 and 60.0. It was 2.0 in my previous change because I kept the "digit.digits" format ... As we have discussed, this is then corrected in the to_endf6 method which returns the value in the "digit.digits" format. If you prefer, I can correct the value of rtyp here instead of in the to_endf6 method. It depends if the _write_rdd(sec) method must have the same syntax of the section mf=8 mt=457 of the endf6 file or it must keep the information written as in sec

luca-fiorito-11: yes, it must be changed in  `_write_rdd`. There is no reason for it to have a different syntax in `_write_rdd` and `to_endf6`.
What is important is consistency! if you change it in `_write_rdd` it should be fixed also for `to_endf6` because in the latter you call [write_mf8](https://github.com/EnricaBelfiore/sandy/blob/BR/sandy/decay.py#L772 ).

Also there is a critical problem if the implemented tests do not fail when they arewrong, right @AitorBengoechea ?!
Tests **MUST NOT** be changed to pass if they are not correct!

AitorBengoechea:
Enrica, Luca has merge two branches into develop. Update the branch from upstream to avoid conflict with the base branch

luca-fiorito-11 (approved):

PR #120, opened 2022-02-02
Lpc tests

AitorBengoechea:


sandy/core/lpc.py
```diff
@@ -244,6 +245,12 @@ def reshape(self, eg, selected_mat=None, selected_mt=None, inplace=False):
             MT number for which the reshape will apply (all MT by default)
         inplace : `bool`, optional, default is `False`
             flag to activate inplace replacement
+        kind: `str` or `int`, optional
```
luca-fiorito-11, line 248: `kind` should not be present...the interpolation must be linear, as mentioned in the docstring description! this is because for LPC the ENDF-6 format states that the interpolation is linear

sandy/core/lpc.py
```diff
@@ -229,7 +229,8 @@ def custom_perturbation(self, mat, mt, p, pert, inplace=False):
         else:
             return Lpc(u_lpc.data)
 
-    def reshape(self, eg, selected_mat=None, selected_mt=None, inplace=False):
+    def reshape(self, eg, selected_mat=None, selected_mt=None, inplace=False,
```
luca-fiorito-11, line 232: can you remove the `inplace` option. to be consistent with other classes I remove `inplace from all methods`

sandy/shared.py
```diff
@@ -236,6 +236,11 @@ def reshape_differential(x, y, xnew):
         new energy grid
     y : `numpy.ndarray` with at least two entries and same length as `x`
         array to interpolate
+    kind: `str` or `int`, optional
```
luca-fiorito-11, line 239: as for the comment in LPC. this is only for linear interpolation

luca-fiorito-11 (changes requested):

luca-fiorito-11 (approved):

luca-fiorito-11:
@AitorBengoechea, thanks for all the tests!

PR #122, opened 2022-02-04
Covariance

AitorBengoechea:


AitorBengoechea:
Dont pay attention to fy.py file. I change to pass the test

sandy/core/cov.py
```diff
@@ -786,48 +762,40 @@ def _gls_G(self, S, Vy_extra=None, sparse=False):
         0	6.00000e+00	1.10000e+01
         1	1.10000e+01	2.60000e+01
 
-        >>> S = np.array([[1, 2], [3, 4]])
-        >>> sensitivity = sandy.CategoryCov.from_var([1, 1])
-        >>> Vy = np.diag(pd.Series([1, 1]))
-        >>> sensitivity._gls_G(S, Vy, sparse=True)
+        >>> sensitivity._gls_G(S, Vy, rows=1)
```
luca-fiorito-11, line 765: don't call it `sensitivity`...it is a covariance matrix

sandy/core/cov.py
```diff
@@ -1815,7 +1849,7 @@ def from_list(cls, iterable):
         return cls(matrix, index=index, columns=index)
 
 
-def corr2cov(corr, s):
+def corr2cov(corr, s, rows=None):
```
luca-fiorito-11, line 1852: I don't want rows here!

sandy/gls.py
```diff
@@ -27,7 +28,7 @@
 
 
 def gls_update(x_prior, S, Vx_prior, Vy_extra, y_extra, sparse=False,
```
luca-fiorito-11, line 30: sparse should be gone!

sandy/core/cov.py
```diff
@@ -1844,30 +1881,229 @@ def corr2cov(corr, s):
     2  12   0  45  0
     3   0   0   0  0
 
+    >>> corr2cov(var, S, rows=1)
```
luca-fiorito-11, line 1884: the test must be corrected

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -306,41 +306,13 @@ def eig(self, sparse=False, sort=True):
...
-                E, V = scipy.linalg.eig(cov.toarray())
-        else:
-            cov = self.data
-            E, V = scipy.linalg.eig(cov)
+        """
+        cov = self.to_sparse()
+        warnings.filterwarnings("ignore")
+        try:
+            E, V = spsl.eigs(cov)
+        except TypeError or ValueError:
```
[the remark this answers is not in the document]
AitorBengoechea: Is not always posible to use sparse.linalg module. Some times gives TypeError or ValueError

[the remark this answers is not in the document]
luca-fiorito-11: I don't agree...not according to this https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigs.html

And even if it were true  then why `scipy.linalg.eig` would work?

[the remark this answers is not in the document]
luca-fiorito-11: I don't agree...not according to https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigs.html

And even if it were true why scipy.linalg.eig would work?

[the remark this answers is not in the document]
AitorBengoechea: The problem with this method is that a very large sparcity is needed for it to work. For example, the test arrays give the following error: TypeError: Cannot use scipy.linalg.eig for sparse A with k >= N - 1. Use scipy.linalg.eig(A.toarray()) or reduce k.

That's why I put a try, so that small arrays are done with scipy.linalg and large arrays are done with scipy.sparse.linalg.

sandy/core/cov.py
```diff
@@ -272,6 +272,10 @@ def data(self, data):
         self._data = pd.DataFrame(data, dtype=float)
         if not len(data.shape) == 2 and data.shape[0] == data.shape[1]:
             raise TypeError("covariance matrix must have two dimensions")
+        if (np.diag(data) >= 0).all() is False:
```
luca-fiorito-11, line 275: can you test it?

sandy/core/cov.py
```diff
@@ -1245,17 +1245,23 @@ def corr2cov(cls, corr, std, rows=None, **kwargs):
             square 2D correlation matrix
         std : 1d `numpy.ndarray`
             array of standard deviations
```
luca-fiorito-11, line 1247: define **kwargs in docstring

luca-fiorito-11 (changes requested):

PR #127, opened 2022-02-14
Correlations

AitorBengoechea:


sandy/core/cov.py
```diff
@@ -283,12 +284,15 @@ def data(self):
...
-            raise TypeError("covariance matrix must have two dimensions")
+            raise TypeError("Covariance matrix must have two dimensions")
         if not (np.diag(data) >= 0).all():
-            raise TypeError("covariance matrix must have positive variance")
+            raise TypeError("Covariance matrix must have positive variance")
         # Round to avoid numerical fluctuations
         if not (data.values.round(14) == data.values.T.round(14)).all():
-            raise TypeError("covariance matrix must be symmetric")
+            raise TypeError("Covariance matrix must be symmetric")
+        correlation = sps.triu(sps.csr_matrix(data.values), k=1)
```
luca-fiorito-11, line 293: This is wrong! Think about it and correct it

sandy/core/cov.py
```diff
@@ -557,57 +565,28 @@ def sampling(self, nsmp, seed=None):
...
-        2  4.84565e-01 -9.91209e-02
+        0  1.74945e+00 -1.73202e+00
+        1 -2.86073e-01 -1.22022e-01
+        2 -4.84565e-01 -4.86773e-01
+
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, rows=1)
+                     0            1
+        0  1.74945e+00 -1.73202e+00
+        1 -2.86073e-01 -1.22022e-01
+        2 -4.84565e-01 -4.86773e-01
```
luca-fiorito-11, line 576: To prove that these samples are correct, can you add a test when you sample from the same 2×2 covariance matrix (same seed). This time take 1 million samples. The resulting covariance (calculated with no.cov) should be equal to the original one within a tolerance limit. Also, the mean of the samples should be 0 within a tolerance limit.

luca-fiorito-11 (changes requested):

luca-fiorito-11:
@EnricaBelfiore , please follow this PR.

luca-fiorito-11:
@AitorBengoechea , we should discuss the whole cholesky approximation method before merging

sandy/core/cov.py
```diff
@@ -1576,33 +1574,32 @@ def _cholesky_aproximation(self, rows=None):
...
-        >>> sandy.CategoryCov([[1, -2],[-2, 3]])._cholesky_aproximation(rows=2).round(3)
+        >>> sandy.CategoryCov([[1, -2],[-2, 3]])._to_positive(rows=2).round(3)
         array([[ 1.171, -1.894],
                [-1.894,  3.065]])
         """
         E, V = self.eig(sort=False)
         if (E >= 0).all():
             positive_matrix = self.data.values
         else:
+            logging.warning('Relative importance of changed eigenvalues:'
```
luca-fiorito-11, line 1594: just the largest in absolute value, not all

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -381,6 +382,8 @@ def corr(self):
             coeff = np.true_divide(1, self.std.values)
             coeff[~ np.isfinite(coeff)] = 0   # -inf inf NaN
         corr = np.multiply(np.multiply(cov, coeff).T, coeff)
+        if not corr.max() <= 1 and corr.min() >= -1:
```
luca-fiorito-11, line 385: @AitorBengoechea , this should be tested already in the data setter...just by calling method corr once data is initialized. Then you would have this warning only at the init level and not every time you run .corr

sandy/core/cov.py
luca-fiorito-11: @AitorBengoechea , this should be kept in mind

luca-fiorito-11 (approved):

PR #129, opened 2022-02-16
CategoryCov unique method for calculation

AitorBengoechea:


AitorBengoechea:
All the calculations in one way, as we talked about but at the time I was not able to visualise it.

sandy/core/cov.py
```diff
@@ -382,7 +382,7 @@ def corr(self):
             coeff = np.true_divide(1, self.std.values)
             coeff[~ np.isfinite(coeff)] = 0   # -inf inf NaN
         corr = np.multiply(np.multiply(cov, coeff).T, coeff)
-        if not corr.max() <= 1 and corr.min() >= -1:
+        if not corr.max() >= 1 and corr.min() <= -1:
```
luca-fiorito-11, line 385: This should be tested. Also, move it to the data setter

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: From the data setter you cannot access a class property without repeating the corr() code.

AitorBengoechea: Maybe, change the `corr` to a method and put a optional argument that link `corr` and `check_correlations` in the following way: 
`corr(self, check=False)`

luca-fiorito-11: good idea! but then i would call it `get_corr`

EnricaBelfiore: I think it is better to have two different methods: in a loop a condition could be `cov.cor() if cov.check=='True'.`

EnricaBelfiore: And I think the warning inserted appears in some cases in which the covariance matrix is not positive defined but also for some "truncation errors". I mean when we perform the multiplication with np.multiply (i.e. the division by the std) we can obtain in the diagonal `1.0000000000000002` instead of `1.0` even with a positive defined matrix

luca-fiorito-11: ok, let's forget about checking the correlations. Anyways, if the eigenvalues are non-negative, then correlations shall always be within -1 and 1.
@AitorBengoechea will check it on a 2x2, 3x3 and4x4 matrix analytically.

The PR can be finalized without the check_correlation option

luca-fiorito-11 (approved):

PR #138, opened 2022-02-18
eig no sparse

EnricaBelfiore:
There is no way with the `scipy.sparse.linalg.eigs()` to obtain all the eigenvalues. The parameter k must be smaller than N-1, with N the dimension of the square matrix. I removed the option of doing it with the sparse matrix and now it is done with `scipy.linalg.eig()`. I put another example with a real covariance matrix

### DIFF
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -324,13 +324,14 @@ class CategoryCov():
                     0            1
         0 7.07107e-01 -7.07107e-01
         1 7.07107e-01  7.07107e-01
+        
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922350)
+        >>> Cov = sandy.XsCov.from_endf6(endf6)
+        >>> Cov = sandy.CategoryCov(Cov)
+        >>> len(Cov.eig()[0])
+        295
         """
-        cov = self.to_sparse()
-        warnings.filterwarnings("ignore")
-        try:
-            E, V = spsl.eigs(cov)
-        except TypeError or ValueError:
-            E, V = scipy.linalg.eig(cov.toarray())
+        E, V = scipy.linalg.eig(self.data)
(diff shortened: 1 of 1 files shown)


luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -324,12 +324,24 @@ def eig(self, sort=True):
...
+        >>> sandy.CategoryCov(Cov).eig()[0]
+        0    8.49942e-01
+        1    1.18850e-01
+        2   -3.32188e-02
+        Name: eigenvalues, dtype: float64
+
+        >>> endf6 = sandy.get_endf6_file("jeff_33", "xs", 10010)
+        >>> err = endf6.get_errorr(ek=sandy.energy_grids.CASMO12, err=1)
+        >>> Cov = err.get_cov()
+        >>> len(Cov.eig()[0]), Cov.data.shape
```
luca-fiorito-11, line 343: I just prefer testing things without relying on on-screen painting if possible.
We can use assert
`assert len(Cov.eig()[0]) == Cov.data.shape[0]`

sandy/core/cov.py
```diff
@@ -324,12 +324,24 @@ def eig(self, sort=True):
...
         >>> Cov = sandy.CategoryCov(Cov)
         >>> len(Cov.eig()[0])
-        295
+        50
+
+        >>> Cov = sandy.random_cov(3,seed=1)
+        >>> sandy.CategoryCov(Cov).eig()[0]
+        0    8.49942e-01
+        1    1.18850e-01
+        2   -3.32188e-02
```
luca-fiorito-11, line 337: I am troubled by this negative eigenvalue...is it correct? Is the `random_cov` method correct?

luca-fiorito-11 (approved):

PR #139, opened 2022-02-19
random correction

EnricaBelfiore:
The random_corr method implemented did not work for matrices with a dimention higher than 30x30 because it returned a matrix not always positive definite nor symettrical (for numerical fluctuations). With this correction the matrix will allow for the creation of a category cov instance up until size 10000x10000 (bigger matrix gives memory issues, in the previous implementation as well), but some eigenvalues could be `-1e-18` or in this range (numerical approximation of zero).  

The method  _reduce_size is modified because it did not work for a real covariance matrix. 

P.s. even if we will change the decomposition it should be mantained because it is used also for different methods (invert, _gls_G_inv)

### DIFF
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -307,6 +307,7 @@ class CategoryCov():
         tolerance : `float`, optional, default is `None`
             truncation proportional to the magnitude of the largest positive eigenvalue,
             e.g. 1000 to cut all the eigenvalues 1000 times lower than the largest eigenvalue
+            or np.inf to cut all the eigenvalues lower than zero
 
         Returns
         -------
@@ -330,6 +331,16 @@ class CategoryCov():
         
         >>> sandy.CategoryCov([[0.1, 0.1], [0.1, 1]]).eig(tolerance=10)[0]
         0   1.01098e+00
+        1   0.00000e+00
+        Name: eigenvalues, dtype: float64
+        
+        >>> sandy.CategoryCov([[1, 2], [2, 1]]).eig()[0]
+        0    3.00000e+00
+        1   -1.00000e+00
+        Name: eigenvalues, dtype: float64
+        
+        >>> sandy.CategoryCov([[1, 2], [2, 1]]).eig(tolerance=np.inf)[0]
+        0   3.00000e+00
         1   0.00000e+00
         Name: eigenvalues, dtype: float64
 
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -327,21 +327,19 @@ class CategoryCov():
 
         >>> Cov = sandy.random_cov(50,seed=11)
         >>> Cov = sandy.CategoryCov(Cov)
-        >>> len(Cov.eig()[0])
-        50
+        >>> assert len(Cov.eig()[0]) == Cov.data.shape[0]
 
         >>> Cov = sandy.random_cov(3,seed=1)
         >>> sandy.CategoryCov(Cov).eig()[0]
-        0    8.49942e-01
-        1    1.18850e-01
-        2   -3.32188e-02
+        0    1.33823e-01
+        1    2.08573e-02
+        2   -6.24761e-19
         Name: eigenvalues, dtype: float64
 
         >>> endf6 = sandy.get_endf6_file("jeff_33", "xs", 10010)
         >>> err = endf6.get_errorr(ek=sandy.energy_grids.CASMO12, err=1)
         >>> Cov = err.get_cov()
-        >>> len(Cov.eig()[0]), Cov.data.shape
-        (39, (39, 39))
+        >>> assert len(Cov.eig()[0]) == Cov.data.shape[0]
         """
         E, V = scipy.linalg.eig(self.data)
         E = pd.Series(E.real, name="eigenvalues")
@@ -399,9 +397,15 @@ class CategoryCov():
                             columns=self.data.columns,
                             )
 
-    def _reduce_size(self):
+    def _reduce_size(self, toll=1e-14):
         """
-        Reduces the size of the matrix, erasing the null values.
+        Reduces the size of the matrix, erasing the zero values.
+
+        Parameters
+        ----------
+        toll : `float`
+            threshold beyond which the eigenvalues are considered to be zero,
+            default is 1e-14
 
         Returns
         -------
@@ -433,7 +437,8 @@ class CategoryCov():
         1	2.00000e+00	0.00000e+00
         2	0.00000e+00	3.00000e+00
         """
-        nonzero_idxs = np.flatnonzero(np.diag(self.data))
+        E = self.eig(sort=False)[0]
+        nonzero_idxs = np.where(abs(E) > toll)[0]
         cov_reduced = self.data.loc[nonzero_idxs, nonzero_idxs]
         return nonzero_idxs, cov_reduced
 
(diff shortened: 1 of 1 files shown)


luca-fiorito-11:
Thanks @EnricaBelfiore , I will have a look. But first I'll wait that you finish the tests

sandy/core/cov.py
```diff
@@ -399,9 +397,15 @@ def get_corr(self):
...
 
-    def _reduce_size(self):
+    def _reduce_size(self, toll=1e-14):
         """
-        Reduces the size of the matrix, erasing the null values.
+        Reduces the size of the matrix, erasing the zero values.
+
+        Parameters
+        ----------
+        toll : `float`
```
luca-fiorito-11, line 406: Better call it `tolerance` than `toll`

sandy/core/cov.py
```diff
@@ -1444,17 +1449,34 @@ def random_corr(cls, size, correlations=True, seed=None, **kwargs):
...
-        1 5.97134e-03 8.52642e-03
+        0 4.59962e-02 4.58318e-02
+        1 4.58318e-02 4.56679e-02
         """
-        corr = random_corr(size, correlations=correlations, seed=seed)
-        std = np.random.uniform(stdmin, stdmax, size)
-        return cls.corr2cov(corr, std, **kwargs)
+        np.random.seed(seed)
+        matrix = scipy.random.rand(size, size)
+        cov = np.cov(matrix)
```
luca-fiorito-11, line 1478: This is wrong, please check what np.cov does!

sandy/core/cov.py
```diff
@@ -2412,10 +2434,11 @@ def random_corr(size, correlations=True, seed=None):
...
 
 
-def random_cov(size, stdmin=0.0, stdmax=1.0, correlations=True, seed=None):
-    corr = random_corr(size, correlations=correlations, seed=seed)
-    std = np.random.uniform(stdmin, stdmax, size)
-    return corr2cov(corr, std)
+def random_cov(size, seed=None):
+    np.random.seed(seed)
+    matrix = scipy.random.rand(size, size)
+    cov = np.cov(matrix)
```
luca-fiorito-11, line 2440: Same as above

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -429,12 +427,12 @@ def _reduce_size(self):
...
-        1	2.00000e+00	0.00000e+00
-        2	0.00000e+00	3.00000e+00
+                    1           2
+        1 2.00000e+00 0.00000e+00
+        2 0.00000e+00 3.00000e+00
         """
         nonzero_idxs = np.flatnonzero(np.diag(self.data))
-        cov_reduced = self.data.loc[nonzero_idxs, nonzero_idxs]
+        cov_reduced = self.data.iloc[nonzero_idxs, nonzero_idxs]
         return nonzero_idxs, cov_reduced
```
luca-fiorito-11, line 436: If you changed this it means that you found an example where `loc` fails and you decided that `iloc` should be used.
Can you add this test/example?

sandy/core/cov.py
```diff
@@ -1444,17 +1442,34 @@ def random_corr(cls, size, correlations=True, seed=None, **kwargs):
...
-        0 2.15373e-02 5.97134e-03
-        1 5.97134e-03 8.52642e-03
+        0 4.59962e-02 4.58318e-02
+        1 4.58318e-02 4.56679e-02
         """
-        corr = random_corr(size, correlations=correlations, seed=seed)
-        std = np.random.uniform(stdmin, stdmax, size)
-        return cls.corr2cov(corr, std, **kwargs)
+        np.random.seed(seed)
+        matrix = scipy.random.rand(size, size)
```
luca-fiorito-11, line 1470: I understand the concept, but then how do you decide the size of the sample.
Have a good look at np.cov to make sure you really know what you are doing

sandy/core/cov.py
```diff
@@ -1444,17 +1442,34 @@ def random_corr(cls, size, correlations=True, seed=None, **kwargs):
         return cls(corr, **kwargs)
 
     @classmethod
-    def random_cov(cls, size, stdmin=0.0, stdmax=1.0, correlations=True,
-                   seed=None, **kwargs):
+    def random_cov(cls, size, seed=None):
         """
+        Construct a covariance matrix with random values
```
luca-fiorito-11, line 1447: I guess we should say that the covariance has unit standard deviation

luca-fiorito-11: And maybe test it...But are we still talking of random_cov or random_corr?

luca-fiorito-11 (changes requested):

sandy/core/cov.py
EnricaBelfiore: The problem is that iloc identifies the position by integer index, while loc uses colunm label-based locationing and, according with the documentation, np.flatnonzero will return the _indices_ of the non-zero elements. This method with `loc` would work only if the columns labels were the same of the indexes (integers in range(df.shape[1])). I saw this problem performing a test with `Cov = sandy.XsCov.from_endf6(endf6)` in which the indeces don't coincide with the columns labels

luca-fiorito-11: Ok! Then add an example that fails with loc

EnricaBelfiore:
I made these changes without inserting the negative eigenvalues equal to zero because this is done by the to_positive method. I prefer to drop that method and insert the cut here if you agree.

luca-fiorito-11:
> I made these changes without inserting the negative eigenvalues equal to zero because this is done by the to_positive method. I prefer to drop that method and insert the cut here if you agree.

We agreed the eig method will remove eigenvalues given a tolerance.
The method `to_semipositive` will reconstruct the covariance with negative eigenvalues set to zero.

sandy/core/cov.py
```diff
@@ -295,14 +296,17 @@ def data(self, data):
...
         """
         Extract eigenvalues and eigenvectors.
 
         Parameters
         ----------
         sort : `bool`, optional, default is `True`
             flag to return sorted eigenvalues and eigenfunctions
+        tolerance : `float`, optional, default is `None`
+            truncation proportional to the magnitude of the largest positive eigenvalue,
+            e.g. 1000 to cut all the eigenvalues 1000 times lower than the largest eigenvalue
```
luca-fiorito-11, line 310: What if you want to remove only negative eigs? What can the user do?

sandy/core/cov.py
```diff
@@ -295,14 +296,18 @@ def data(self, data):
     def size(self):
         return self.data.values.shape[0]
 
-    def eig(self, sort=True):
+    def eig(self, sort=True, tolerance=None):
```
luca-fiorito-11, line 299: maybe we should remove the `sort` option since it is not tested.
It is anyway a very simple implementation that the user only needs for visualization purposes

luca-fiorito-11 (changes requested):

sandy/core/cov.py
```diff
@@ -307,6 +307,7 @@ def eig(self, sort=True, tolerance=None):
         tolerance : `float`, optional, default is `None`
             truncation proportional to the magnitude of the largest positive eigenvalue,
             e.g. 1000 to cut all the eigenvalues 1000 times lower than the largest eigenvalue
+            or np.inf to cut all the eigenvalues lower than zero
```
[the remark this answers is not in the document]
luca-fiorito-11: I did it!

sandy/core/cov.py
luca-fiorito-11: I removed it

luca-fiorito-11 (approved):

PR #140, opened 2022-02-21
tsl files readable

AitorBengoechea:


luca-fiorito-11:
@EnricaBelfiore , can you review this PR?

EnricaBelfiore:
Should you change the `zam` part in the docstring? @AitorBengoechea for nfpy, decay, xs, tsl we can put the “commas” in the same way

luca-fiorito-11:
> Should you change the `zam` part in the docstring? @AitorBengoechea for nfpy, decay, xs, tsl we can put the “commas” in the same way and insert the tsl option in kind

Can you add your review?

EnricaBelfiore (changes requested):
Should you change the `zam` part in the docstring? for nfpy, decay, xs, tsl we can put the “commas” in the same way and insert the `tsl` option in kind

AitorBengoechea:
The tls option is already a option in kind (line 297). Normally in all the other options the zam identifier links the web page adress with the specific file that we want to analyse. In the case of TLS, i used the nomenclature available in the NJOY manual.

EnricaBelfiore:
> The tls option is already a option in kind (line 297). Normally in all the other options the zam identifier links the web page adress with the specific file that we want to analyse. In the case of TLS, i used the nomenclature available in the NJOY manual.

I know, I meant that you should insert it in the `kind` section of the docstring and indeed the zam nomenclature used in tls case

sandy/libraries.py
```diff
@@ -14661,3 +14661,115 @@
     1102731: "decay_3851_110-Ds-273M.dat",
     1112720: "decay_3852_111-Rg-272.dat",
 }
+
+# Thermal Neutron Scattering Data for ENDF-B-VII.1 from the IAEA website
+URL_TSL_ENDFB_71_IAEA = "https://www-nds.iaea.org/public/download-endf/ENDF-B-VII.1/tsl/"
+TSL_FILES_ENDFB_71_IAEA = {
+    1: "tsl_0001_H(H2O).dat",
```
luca-fiorito-11, line 14668: Unfortunately the MAT numbers provided at page 673 of the LEAPR section  of the NJOY manual are not meaningful.
I think we cannot  anymore use integers (like the ZAM number) to identify the compounds. We need real names with strings.

I propose using those reported in the same page of the NJOY manual.
Example: `water`, `BeO`, `Be_in_BeO`.

Probably we need a function that tells the users what keys are available for TSL data, otherwise it would be too complicated to find the right name and match the correct case-sensitive key.

sandy/core/endf6.py
```diff
@@ -170,6 +183,22 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
+
+    Thermal Neutron Scattering Data from ENDF/B-VIII.0. ('all' is tested also)
+    >>> tape = sandy.get_endf6_file("endfb_80", 'tsl', [1, 2, 3])
+    >>> assert type(tape) is sandy.Endf6
+
+    Thermal Neutron Scattering Data from JEFF-3.3. ('all' is tested also)
+    >>> tape = sandy.get_endf6_file("jeff_33", 'tsl', [1, 2, 3])
+    >>> assert type(tape) is sandy.Endf6
+
+    Thermal Neutron Scattering Data from JENDL-4.0u ('all' is tested also)
```
luca-fiorito-11, line 198: To say that it is tested doesn't mean it is. Just create a notebook for 1 library and upload it on the sandy notebook environment.

luca-fiorito-11 (changes requested):

sandy/core/endf6.py
AitorBengoechea: i dont implement the reading of tsl, only the libraries and  the get_endf6_file method

luca-fiorito-11: Sure, for that you would need to update `read_section`

sandy/libraries.py
AitorBengoechea: In my opinion that only makes it more difficult to write and use the code.I agree with both Enrica and Luca that we need something that relates the `zam` of the method to the keys available for TSL. So my question is the following:

If we create a function that gives you everything: the keys and their meaning, why don't we call them with integers? After all, if such a function exists, the key would be decided by us and the user would know it. 

My opinion is the following: Appear in the docstring the whole list (as Enrica said), and change line 326 so that in the case of tsl it prints the whole dictionary with the raise error and in the other cases, it returns the closest values that are available in the library.

AitorBengoechea:
I have created a function that returns the index that is available in all the libraries. It seems to me that this is the easiest way for the potential user to relate the dictionary keys with the information he wants to look up. 
I have also put a note where I explain it

sandy/core/endf6.py
```diff
@@ -56,6 +65,86 @@
...
+            * `'endfb_71'`
+            * `'jeff_33'`
+            * `'endfb_80'`
+            * `'jendl_40u`
+
+    Raises
+    ------
+    ValueError
+        if library is not among available selection.
+
```
EnricaBelfiore, line 86: You forgot the `Return`

sandy/core/endf6.py
```diff
@@ -81,6 +170,11 @@ def get_endf6_file(library, kind, zam, to_file=False):
...
             * `'jeff_33'`
             * `'endfb_80'`
+        for 'tsl' (read the note)
+            * `'endfb_71'`
+            * `'jeff_33'`
+            * `'endfb_80'`
+            * `'jendl_40u`
     kind : `str`
         nuclear data type:
             * `xs` is a standard neutron-induced nuclear data file
```
EnricaBelfiore, line 180: You should add *`tsl`

EnricaBelfiore (changes requested):

sandy/core/endf6.py
AitorBengoechea: There is no return, only print data on the screen

sandy/core/endf6.py
```diff
@@ -56,6 +65,86 @@
...
+    user_agent = 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.9.0.7) Gecko/2009021910 Firefox/3.0.7'
+    headers = {'User-Agent': user_agent, }
+    request = Request(index, None, headers)
+    response = urlopen(request)
+    data = response.read().decode("utf-8")
+    # Remove html style:
+    data = data[data.find('<pre>')+5:data.find('</pre>')]
+    data = re.sub(r'">tsl\S+', '', data)
+    data = re.sub(r'<a href="tsl/', '', data)
+    print(data.replace('MAT', 'KEY'))
```
EnricaBelfiore, line 144: Can we call it _index_ instead of _key_? To be consistent with line 70 or we should change line 70

EnricaBelfiore: I understand, but it's not so clear for me that you need to put the `key` value in the` zam` section of the following function. I meant how can I know that the correct integer for graphite (line 105) is 31 and not 10? Maybe we should change the note or do we have to assume that the user knows this?

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+        "read_mf7",
+        "write_mf7",
+        ]
+
+mf = 7
+
+
+def read_mf7(tape, mat, mt):
+    """
+    Parse MAT/MF=MAT/8 section from `sandy.Endf6` object and return
```
EnricaBelfiore, line 31: Should it be `MAT/MF=7/MT`?

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+    -------
+    `str`
+        Multiline string reproducing the content of a ENDF-6 section.
+
+    Examples
+    --------
+    Incoherent inelastic scattering:
+    >>> tls = sandy.get_endf6_file("endfb_80", 'tsl', 26)
+    >>> sec = sandy.sections.mf7._read_incoherent_inelastic(tls, 26, 4)
+    >>> assert len(_write_inelastic_scattering(sec)) == len(tls.data[(26, 7, 4)])
```
EnricaBelfiore, line 395: Can you check why this example cannot work for mat = 1 or 11, 40, 47, 49, 51?

AitorBengoechea: I will check, thanks

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+
+    Examples
+    --------
+    Coherent elastic scattering:
+    >>> tls = sandy.get_endf6_file("endfb_80", 'tsl', 26)
+    >>> sec = sandy.sections.mf7._read_elastic_scattering(tls, 26, 2)
+    >>> assert len(_write_elastic_scattering(sec)) == len(tls.data[(26, 7, 2)])
+
+    This test is also check:
+    >>> #assert print(_write_elastic_scattering(sec)) == print(tls.data[(26, 7, 2)])
```
luca-fiorito-11, line 324: either you keep it or you remove it

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+
+def _write_inelastic_scattering(sec):
+    """
+    Given the content of MF7 for inelastic scattering as nested dictionaries,
+    write it to string.
+
+    Parameters
+    ----------
+    sec : 'dict'
+        Multiline string reproducing the content of a ENDF-6 section.
```
luca-fiorito-11, line 383: rewrite

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+    `str`
+        Multiline string reproducing the content of a ENDF-6 section.
+
+    Examples
+    --------
+    Incoherent inelastic scattering:
+    >>> tls = sandy.get_endf6_file("endfb_80", 'tsl', 26)
+    >>> sec = sandy.sections.mf7._read_incoherent_inelastic(tls, 26, 4)
+    >>> assert len(_write_inelastic_scattering(sec)) == len(tls.data[(26, 7, 4)])
+
```
luca-fiorito-11, line 396: we should have a test that compares the entire string

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+    This test is also check: 
+    >>> # assert print(_write_inelastic_scattering(sec)) == print(tls.data[(26, 7, 4)])
```
luca-fiorito-11, line 398: either it stays or it goes

sandy/tsl.py
```diff
@@ -0,0 +1,222 @@
...
+        instance.
+
+        Parameters
+        ----------
+        tape : `sandy.Endf6`
+            instance containing tsl data
+
+        Returns
+        -------
+        `dict`
```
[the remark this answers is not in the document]
luca-fiorito-11: correct the docstring

luca-fiorito-11 (changes requested):

sandy/sections/mf7.py
EnricaBelfiore: I think that this example must be removed or changed. Even if you do `print(1) == print(3)` the result will be `True` because print returns _None_, indeed `print(1) == None`.

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+
+    * `read_mf7`
+    * `write_mf7`
+
+Function `read` reads a MF7/MT section from a string and produces a content
+object with a dictionary-like structure.
+The content object can be accessed using most of the keywords specified in
+the ENDF6 manual for this specific MF section.
+
+Function `write_mf8` writes a content object for a MF7/MT section into a
```
EnricaBelfiore, line 13: `write_mf7`

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+        C, i = sandy.read_tab1(df, i)
+        temp = C.C1  # Temperature
+        LT = C.L1  # Temperature flag
+        add_2 = {
+            "LT": LT,
+            "NBT": C.NBT,  # Number of different pairs alpha S
+            "INT": C.INT,
+            "S": C.y,
+            }
+        add["EINT"] = C.x  # Energy array, constant
```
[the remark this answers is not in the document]
AitorBengoechea: Because i am centering all the update in the last line.

sandy/sections/mf7.py
```diff
@@ -0,0 +1,448 @@
...
+    out.update(add)
+    if B[0] != 0:
+        C, i = sandy.read_tab2(df, i)
+        add = {
+                'NR': C.NZ,
+                'NP': C.NBT,
+                'BINT': C.INT,
+        }
+        Num_beta = C.NBT[0]
+        add_beta = {}
```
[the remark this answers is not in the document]
AitorBengoechea: I have problems creating two levels in a row in `dict`. If you do `add[b][temp]` you will recive a keyError

sandy/core/endf6.py
```diff
@@ -170,6 +271,22 @@ def get_endf6_file(library, kind, zam, to_file=False):
     >>> tape = sandy.get_endf6_file("jeff_33", 'decay', [380900, 551370, 541350])
     >>> assert type(tape) is sandy.Endf6
 
+    Thermal Neutron Scattering Data from ENDF/B-VII.1. ('all' is tested also)
```
luca-fiorito-11, line 274: Remove "all is tested also".

luca-fiorito-11 (changes requested):
I didn't review tsl.py

luca-fiorito-11 (approved):

Issue #137, opened 2022-02-18
Cholesky decomposition


luca-fiorito-11:
### Discussed in https://github.com/luca-fiorito-11/sandy/discussions/133

<div type='discussions-op-text'>

<sup>Originally posted by **EnricaBelfiore** February 18, 2022</sup>
@luca-fiorito-11 @AitorBengoechea 

I and @GrimFe noticed that the last modification introduced in the sampling procedure does not work for all nuclear data covariance matrices. More in the details, we tested it for covariances of Pu239, U235, U233 and H1.
First of all, was this method already tested with real test cases? In case it was, which nuclides did you use for testing @AitorBengoechea?

We think that this problem is related to the fact that the covarinace for nuclear data are not always positive defined. As far as we understand, `to_positive` method handles this by transforming such matrices into SEMI-positive defined matrices. LU decomposition implemented inside `sparse_table_cholesky` method does not properly work for singular matrix (det=0).
 
We propose to back up to the previous decomposition scheme (**cov = U*D*U.T**) and to return **L=U*sqrt(D)**, so that **cov=L*L.T**, where cov is the semipositive approximation of the nuclear data covariance matrix. 
Unfortunately, `scipy.linalg.ldl` (implementing UDU.T decomposition) does not handle sparse matrix. Therefore our suggestion is to directly implement `scipy.linalg.ldl` decomposition in `get_L`. Further possible memory optimization can and should be considered. 
</div>

luca-fiorito-11:
Close with PR #143

PR #143, opened 2022-02-25
Decompose

AitorBengoechea:


luca-fiorito-11:
why did you move `_reduce_size` and `_restore_size` outside `CategoryCov`?

luca-fiorito-11:
Is it because you want to use these methods for the `L` matrix?

luca-fiorito-11:
> Because `_restore_size` returns a `CategoryCov` object, and lower triangular matrix is not a covariance matrix

then, do we need to keep them in `CategoryCov`?
Where else are they used?

AitorBengoechea:
In line 1637 you obtain the `L_reduce`. This mean that `L_reduce`.dot(`L_reduce.T`) = `cov_reduce`. This is not what `get_L` should return. The objective of the method is given a tolerance for eigenvalues and a `threshold`, return a the L matrix that do the following `L.dot(L) = cov`. The reason of reducing size is compute easier the calculation of L, becuase rows and columns fully of zeros doesnt affect the final result.
If you want to do some approximattion in the samples and working with reduce size, it should go in another method.

AitorBengoechea:
The method `invert` uses, but only for obtaining the `pd.DataFrame`

luca-fiorito-11:
> d is given a tolerance for eigenvalues and a `threshold`, return a the L matrix that do the following `L.dot(L) = cov`. The reason of reducing size is compute easier the calculation of L, becuase rows and columns fully of zeros doesnt affect the final result.
> If you want to do some approximattion in the samples and working with reduce size, it should go in another method.

ok, but this does not answer the question. Do we need to keep them in `CategoryCov`?

AitorBengoechea:
In my opinion, no. They are general functions that helps in the calculations.

luca-fiorito-11:
> In my opinion, no. They are general functions that helps in the calculations.

Then I agree to remove them from `CategoryCov`.
Can you make sure that the existing tests and docstrings are also copied outside `CategoryCov`?...

AitorBengoechea:
> > In my opinion, no. They are general functions that helps in the calculations.
> 
> Then I agree to remove them from `CategoryCov`. Can you make sure that the existing tests and docstrings are also copied outside `CategoryCov`?...
They are a copy of the ones that are in `CategoryCov`, also i adapt the tests to obtain the same results of the `CategoryCov` methods

sandy/core/cov.py
```diff
@@ -1605,87 +1531,30 @@ def get_L(self, rows=None):
...
+        E = sandy.CategoryCov.from_var(np.sqrt(E)).data.values
+        # Construct the matrix:
+        rows_ = cov_reduced.shape[0] if rows is None else rows
+        A = sandy.cov.sparse_tables_dot(V, E, rows=rows_).T.toarray()
+        # QR decomposition:
+        Q, R = scipy.linalg.qr(A)
+        L_redu = R.T
+        # Original size
+        L = restore_size(nonzero_idxs, L_redu, len(self.data)).values
+        if threshold is not None:
```
luca-fiorito-11, line 1554: I am not sure we need a threshold...are we ever going to use it?
We will not implement it in the sampling pipeline.
And if it is just "in case", then any API user can make this simple truncation...

sandy/core/cov.py
```diff
@@ -1605,87 +1531,30 @@ def get_L(self, rows=None):
...
-        1	-1.75078e+00	0.00000e+00
-
-        >>> sandy.CategoryCov([[1, -2],[-2, 3]]).get_L(rows=1)
-                       0   	          1
-        0	 1.08204e+00	0.00000e+00
-        1	-1.75078e+00	0.00000e+00
+                       0	          1	          2
+        0	-2.00000e+00	0.00000e+00	0.00000e+00
+        1	-6.00000e+00	1.00000e+00	0.00000e+00
+        2	 8.00000e+00	5.00000e+00	3.00000e+00
```
luca-fiorito-11, line 1537: Can we have a test where L L^T =C?

sandy/core/cov.py
```diff
@@ -2480,6 +2301,92 @@ def sample_distribution(dim, nsmp, seed=None, pdf='normal'):
...
+    --------
+    >>> S = pd.DataFrame(np.diag(np.array([0, 2, 3, 0])))
+    >>> M_nonzero_idxs, M_reduce = reduce_size(S)
+    >>> M_reduce[::] = 1
+    >>> restore_size(M_nonzero_idxs, M_reduce.values, len(S))
+                0           1           2           3
+    0 0.00000e+00 0.00000e+00 0.00000e+00 0.00000e+00
+    1 0.00000e+00 1.00000e+00 1.00000e+00 0.00000e+00
+    2 0.00000e+00 1.00000e+00 1.00000e+00 0.00000e+00
+    3 0.00000e+00 0.00000e+00 0.00000e+00 0.00000e+00
```
luca-fiorito-11, line 2381: Can we check that the combination `reduce_size` and then `restore_size ` preserves indices and columns?

luca-fiorito-11 (changes requested):

sandy/core/cov.py
AitorBengoechea: I put a test demostrating that the original index are not kept in the restore part. For cov reduced is kept.

luca-fiorito-11 (approved):

luca-fiorito-11:
Closes issue #137

PR #148, opened 2022-03-01
from 0 to 99999

AitorBengoechea:


AitorBengoechea:
The len problem in the tls.fy read and write is solved

sandy/core/records.py
```diff
@@ -327,9 +327,40 @@ def write_int(x):
...
+    99999
+    >>> np.array(line_numbers(1.0e6+1)).min()
+    1
+    >>> np.array(line_numbers(1.0e4+1)).max()
+    10001.0
+    """
     iend = istart + length
-    ilines = np.arange(istart, iend)
-    return np.where(ilines < 100000, ilines, ilines-99999).tolist()
+    if length < 100000:
```
luca-fiorito-11, line 355: I would do something much simpler, like
`np.tile(np.arange(1, 1e5, dtype=int), int(iend//99999)+1)[:iend]`

I am not even sure we need `istart`

sandy/core/records.py
```diff
@@ -327,9 +327,40 @@ def write_int(x):
...
+        List containing the number of each line.
+
+    Examples
+    --------
+    >>> np.array(line_numbers(1.0e6)).max()
+    99999
+    >>> np.array(line_numbers(1.0e6+1)).min()
+    1
+    >>> np.array(line_numbers(1.0e4+1)).max()
+    10001.0
```
luca-fiorito-11, line 352: I don't like having float numbers while all outputs should be integers

sandy/core/records.py
```diff
@@ -327,9 +327,40 @@ def write_int(x):
...
+        Number from where start counting. The default is 1.
+
+    Returns
+    -------
+    ilines : `list`
+        List containing the number of each line.
+
+    Examples
+    --------
+    >>> np.array(line_numbers(1.0e6)).max()
```
luca-fiorito-11, line 347: these three tests are not enough.
we should at least check for the size...and maybe more

luca-fiorito-11 (changes requested):

sandy/core/records.py
AitorBengoechea: IT is because this specific part is a numpy array. The `(line_numbers(1.0e4+1)` is `int`

sandy/core/records.py
AitorBengoechea: Okay, the real thing is that i test with tls files and now len are the same in all the cases

luca-fiorito-11 (commented):
Still some tests are missing

luca-fiorito-11 (approved):

PR #161, opened 2022-03-14
get_DE

EnricaBelfiore:


sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
...
+        """
+        DecayEnergy = []
+        DecayEnergyUncertainty = []
+        ZAM = []
+        for zam, dic in self.data.items():
+            DecayEnergy.extend([
+                dic['decay_energy']['alpha'],
+                dic['decay_energy']['beta'],
+                dic['decay_energy']['gamma'],
+                dic['decay_energy']['total'],
```
luca-fiorito-11, line 294: We should not use total. This should be reconstructed. Maybe it should even be removed form`DecayData`

sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
...
+                beta    5.06717e+04
+                gamma   1.63616e+05
+                total   4.67889e+06
+        942400  alpha   5.24303e+06
+                beta    1.11164e+04
+                gamma   1.36292e+03
+                total   5.25551e+06
+        Name: BE, dtype: float64
+        """
+        DecayEnergy = []
```
luca-fiorito-11, line 286: No uppercase letters

sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
...
+            DecayEnergyUncertainty.extend([
+                dic['decay_energy_uncertainties']['alpha'],
+                dic['decay_energy_uncertainties']['beta'],
+                dic['decay_energy_uncertainties']['gamma'],
+                dic['decay_energy_uncertainties']['total'],
+                ])
+            ZAM.append(zam)
+        name = ['alpha', 'beta', 'gamma', 'total']
+        df = pd.DataFrame(zip(DecayEnergy, DecayEnergyUncertainty),
+                          index=pd.MultiIndex.from_product([ZAM, name]),
```
luca-fiorito-11, line 305: I am not sure about using a multiindex. Maybe they should just be all column. The same for `get_br`

sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
         else:
             return df.BR
 
+    def get_decay_energy(self, with_uncertainty=True):
```
luca-fiorito-11, line 240: Here you use underscores, while in `get_decay_constants` you don't. Then in `get_br` you use initials. Choose one and be consistent! This is the one I prefer

sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
...
+               gamma 1.63616e+05  1.70801e+03
+               total 4.67889e+06  1.63320e+05
+        942400 alpha 5.24303e+06  3.63881e+04
+               beta  1.11164e+04  9.02572e+02
+               gamma 1.36292e+03  1.33403e+02
+               total 5.25551e+06  3.63995e+04
+
+        >>> endf6 = sandy.get_endf6_file("jeff_33", "decay", [942400, 922350])
+        >>> rdd = sandy.DecayData.from_endf6(endf6)
+        >>> DE = rdd.get_decay_energy(False)
```
luca-fiorito-11, line 274: No capital letters

sandy/decay.py
```diff
@@ -230,6 +237,78 @@ def get_br(self, with_uncertainty=True):
...
+                dic['decay_energy_uncertainties']['alpha'],
+                dic['decay_energy_uncertainties']['beta'],
+                dic['decay_energy_uncertainties']['gamma'],
+                dic['decay_energy_uncertainties']['total'],
+                ])
+            ZAM.append(zam)
+        name = ['alpha', 'beta', 'gamma', 'total']
+        df = pd.DataFrame(zip(DecayEnergy, DecayEnergyUncertainty),
+                          index=pd.MultiIndex.from_product([ZAM, name]),
+                          columns=['BE', 'Uncertainty'])
```
luca-fiorito-11, line 306: Why is it called `BE`? Why here the uncertainty is called "uncertainty" and in `get_br` it is "DBR"? Be consistent! 
I suggest here using "E" and "DE".

sandy/decay.py
luca-fiorito-11: By the way, I guess switching `get_decayconstant` to `get_half_lives` will facilitate the pipeline. What do you think?

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: Indeed we will use the already implemented  `get_half_life` method to extract the BE information so I will remove `get_decay_constant` and sobstitute  the `custom_perturbation_decayconstant` method with the one for half life in another pull request

sandy/decay.py
```diff
@@ -182,55 +183,132 @@ def get_br(self, with_uncertainty=True):
...
+        942410      alpha 2.44000e-05
+        942410       beta 9.99976e-01
+        """
+        br = []
+        zam = []
+        for z, dic in self.data.items():
+            try:
+                for rtyp, dk in dic['decay_modes']:
+                    br.append([
+                        rtyp2decaymode(rtyp),
```
luca-fiorito-11, line 228: I would keep te RTYP like before. What if a nuclide decays beta and emits also 2 neutrons?

luca-fiorito-11 (changes requested):

sandy/decay.py
EnricaBelfiore: It works also for multiple particle decay.
In the case that you mentioned the function will return `'beta, n emission, n emission'`.  In this way you avoid having zero that has no meaning in rtyp. Do you mean it's too long?

luca-fiorito-11: For the moment we don't need this. We can keep the RTYP

sandy/decay.py
```diff
@@ -184,61 +184,61 @@ def get_branching_ratio(self, with_uncertainty=True):
...
+        942410   10 9.99976e-01
         """
         br = []
         zam = []
         for z, dic in self.data.items():
             try:
                 for rtyp, dk in dic['decay_modes']:
                     br.append([
-                        rtyp2decaymode(rtyp),
+                        rtyp,
```
luca-fiorito-11, line 228: Why try/except?  Cannot you use `if "decay_modes" in dic` or `if dic["decay_modes"]`?

luca-fiorito-11 (changes requested):

sandy/decay.py
```diff
@@ -222,15 +222,15 @@ def get_branching_ratio(self, with_uncertainty=True):
...
+            if 'decay_modes' in dic.keys():
                 for rtyp, dk in dic['decay_modes']:
                     br.append([
                         rtyp,
                         dk['branching_ratio'],
                         dk['branching_ratio_uncertainty'],
                         ])
                     zam.append(z)
-            except:
+            else:
```
luca-fiorito-11, line 233: @EnricaBelfiore the warning might have to be captured in a future test

sandy/decay.py
```diff
@@ -139,23 +139,23 @@ def get_halflives(self, with_uncertainty=True):
...
                         HL         DHL
+        ZAM                           
         922350 2.22102e+16 1.57788e+13
         942400 2.07108e+11 1.57785e+08
 
         >>> endf6 = sandy.get_endf6_file("jeff_33", "decay", [942400, 922350])
         >>> rdd = sandy.DecayData.from_endf6(endf6)
-        >>> HL = rdd.get_halflives(False)
-        >>> print(HL)
+        >>> rdd.get_half_life(False)
```
luca-fiorito-11, line 157: @EnricaBelfiore , will need a test for stable nuclides

sandy/decay.py
```diff
@@ -182,55 +183,132 @@ def get_br(self, with_uncertainty=True):
...
         >>> rdd = sandy.DecayData.from_endf6(endf6)
-        >>> BR = rdd.get_br(False)
-        >>> print(BR)
-        ZAM     RTYP
-        922350  40     1.00000e+00
-                60     7.20000e-11
-        942400  40     1.00000e+00
-                60     5.70000e-08
-        Name: BR, dtype: float64
+        >>> rdd.get_decay_energy(False)
```
luca-fiorito-11, line 275: @EnricaBelfiore Need a future test for stable nuclides

sandy/decay.py
```diff
@@ -182,55 +183,132 @@ def get_br(self, with_uncertainty=True):
...
+                dic['decay_energy']['alpha'],
+                dic['decay_energy']['beta'],
+                dic['decay_energy']['gamma'],
+                ])
+            decay_energy_uncertainty.extend([
+                dic['decay_energy_uncertainties']['alpha'],
+                dic['decay_energy_uncertainties']['beta'],
+                dic['decay_energy_uncertainties']['gamma'],
+                ])
+            {zam.append(z) for i in range(3)}
```
luca-fiorito-11, line 300: There are much better ways to do this. To be reimplemented in the future

sandy/decay.py
```diff
@@ -243,22 +321,22 @@ def get_decayconstant(self, with_uncertainty=True):
...
+                   DLAMBDA      LAMBDA
+        ZAM                           
         922350 2.21715e-20 3.12085e-17
 
         >>> endf6 = sandy.get_endf6_file("jeff_33", "decay", 942390)
         >>> rdd = sandy.DecayData.from_endf6(endf6)
-        >>> DC = sandy.DecayData.get_decayconstant(rdd, False)
-        >>> print(DC)
+        >>> rdd.get_decay_constant(False)
+        ZAM
```
luca-fiorito-11, line 339: Need a test for a stable nuclide

luca-fiorito-11 (approved):
Approved! But tests are missing

PR #163, opened 2022-03-16
update to work with errorr in sampling

AitorBengoechea:


sandy/core/endf6.py
```diff
@@ -1744,6 +1744,7 @@ def get_errorr(self,
         1.00000e-05 3.00000e-02 5.80000e-02 1.40000e-01 2.80000e-01 3.50000e-01 6.25000e-01 4.00000e+00 4.80520e+01 5.53000e+03 8.21000e+05 2.23100e+06 1.00000e+07 /
         stop
         """
+        temperature = 0 if temperature is None else temperature
```
luca-fiorito-11, line 1747: I don't understand why `temperature` should be `None`.
This should not be a possibility....if it happens yhere is a problem somewhere else

luca-fiorito-11 (changes requested):

PR #164, opened 2022-03-18
Groupr plus sampling

AitorBengoechea:


luca-fiorito-11 (approved):
@AitorBengoechea , don't forget to add `mat` and `mt` to Parameters

PR #165, opened 2022-03-22
reading update

AitorBengoechea:


sandy/groupr.py
```diff
@@ -147,11 +166,13 @@ def get_xs(self, **kwargs):
                                           listmt=listmt_,
                                           listmat=listmat_).data:
             mf3 = sandy.groupr.read_mf3(self, mat, mt)
-            xs = np.array([x["DATA"][1].tolist() for x in mf3["GROUPS"]]).T
+            lowest_range = mf3["GROUPS"][0]["IG"] - 1
+            xs = np.array([x["DATA"][1].tolist() for x in mf3["GROUPS"]])
+            xs = np.insert(xs, [0]*lowest_range, 0) if lowest_range != 0 else xs
```
luca-fiorito-11, line 171: @AitorBengoechea . This is ok, but pandas bfill could be better, i guess...

sandy/groupr.py
```diff
@@ -197,15 +218,34 @@ def get_flux(self, **kwargs):
         (821000.0, 2231000.0]     1.41000e+06
         (2231000.0, 10000000.0]   7.76900e+06
         Name: iwt, dtype: float64
+
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922350)
+        >>> groupr = endf6.get_gendf(ek=sandy.energy_grids.CASMO12)
+        >>> groupr.get_flux(mt=4)
```
luca-fiorito-11, line 224: Why does the flux change with the MT?

luca-fiorito-11 (changes requested):

sandy/groupr.py
AitorBengoechea: You are right, the flow should not change. It is a mistake to apply the same rule to flux and xs.

sandy/groupr.py
AitorBengoechea: As it was: xs in the array of the values of the xs and columns is the `pd.MultiIndex` and index is the `pd.Index`. For me is more natural continue working with np in the xs variable and them in then combine all in the `data.append`

sandy/groupr.py
luca-fiorito-11: ok, then should we fix the mistake or not?

luca-fiorito-11 (approved):

sandy/groupr.py
luca-fiorito-11: i see it is fixed now

PR #176, opened 2022-04-16
From nuclide to zam

EnricaBelfiore:
Co-authored-by:  GrimFe federico.grimaldi98@gmail.com

Added:

- function `nuclide2zam` to convert string with symbol and mass number to ZAM;
- function `nuclide2za` to convert string with symbol and mass number to ZA;
- function `za2nuclide`, opposite of `nuclide2za`

For the implementation of these points the dictionaries `METASTATE_FLIP`, `ATOMIC_NUMBERS` were introduced


sandy/zam.py
```diff
@@ -567,9 +705,88 @@ def zam2nuclide(zam, atomic_number=False, sep=""):
...
+    zam = nuclide2zam(nuclide)
+    return zam2za(zam)
 
 def get_meta_letter(m, skip_ground=False):
     meta = METASTATES[m]
     if skip_ground and m == 0:
         meta = ""
     return meta
+
+def get_meta_number(m):
```
luca-fiorito-11, line 790: Do we need a function? We already have a dictionary

sandy/zam.py
```diff
@@ -498,6 +627,8 @@ def zam2za(zam, method="nndc"):
 def z2sym(z):
     return ELEMENTS[z]
 
+def sym2z(sym):
+    return ATOMIC_NUMBERS[sym]
```
luca-fiorito-11, line 631: Do we need a function? We already have a dictionary

sandy/zam.py
```diff
@@ -149,6 +271,13 @@
     3: "o",
     }
 
+METASTATES_FLIP = {
```
luca-fiorito-11, line 274: Why not `METASTATES_FLIP = {v: k for k, v in METASTATES.items()}`?

sandy/zam.py
```diff
@@ -141,6 +143,126 @@
     118: 'UUp',
     }
 
+ATOMIC_NUMBERS = {
```
luca-fiorito-11, line 146: Why not `ATOMIC_NUMBERS = {v: k for k, v in ELEMENTS.items()}`

sandy/zam.py
```diff
@@ -567,9 +705,88 @@ def zam2nuclide(zam, atomic_number=False, sep=""):
         out = f"{z}{sep}{out}"
     return out
 
+def za2nuclide(za):
+    zam = za2zam(za)
```
luca-fiorito-11, line 709: And what about all arguments of `za2zam`?

luca-fiorito-11 (changes requested):

sandy/zam.py
EnricaBelfiore: I did it to be consistent with the function `z2sym`, I will remove both

luca-fiorito-11: Do you need them? If yes you keep them, else you remove them.
Maybe also `z2sym` should be removed.

sandy/zam.py
```diff
@@ -698,16 +568,16 @@ def zam2nuclide(zam, atomic_number=False, sep=""):
...
     meta = get_meta_letter(m, skip_ground=True)
     out = f"{sym}{sep}{a}{meta}"
     if atomic_number:
         out = f"{z}{sep}{out}"
     return out
 
-def za2nuclide(za):
-    zam = za2zam(za)
-    return zam2nuclide(zam)
+def za2nuclide(za, method="nndc", meta=0, atomic_number=False, sep=""):
```
luca-fiorito-11, line 578: Docstring and tests?

luca-fiorito-11 (changes requested):

luca-fiorito-11:
@EnricaBelfiore , can we close this PR?

luca-fiorito-11:
> > @EnricaBelfiore , can we close this PR?
> 
> For the moment I have inserted a Note, but I think the `expand_za` function requires more effort. In this moment it cannot discriminate between 'Am242m' and 'Am242n'. If you have any suggestions I'll change it

Something like
```
        if a >= 300 and a < 700:
            m = 1
            a = a - 300 - m*100
        elif a >= 700 and a < 1100:
            m=2
            a=...
        elif a>=1100:
            raise ValueErrorr(ZA jumber is not recognized)
```

EnricaBelfiore:
> > > @EnricaBelfiore , can we close this PR?
> > 
> > 
> > For the moment I have inserted a Note, but I think the `expand_za` function requires more effort. In this moment it cannot discriminate between 'Am242m' and 'Am242n'. If you have any suggestions I'll change it
> 
> Something like
> 
> ```
>         if a >= 300 and a < 700:
>             m = 1
>             a = a - 300 - m*100
>         elif a >= 700 and a < 1100:
>             m=2
>             a=...
>         elif a>=1100:
>             raise ValueErrorr(ZA jumber is not recognized)
> ```

It is not really a "general rule" ... It can be used if the nuclide with more than one isomeric state has a mass number greater than 200, as usually happens, but not for Hf177m and Hf177n for example.

luca-fiorito-11:
> > > > @EnricaBelfiore , can we close this PR?
> > > 
> > > 
> > > For the moment I have inserted a Note, but I think the `expand_za` function requires more effort. In this moment it cannot discriminate between 'Am242m' and 'Am242n'. If you have any suggestions I'll change it
> > 
> > 
> > Something like
> > ```
> >         if a >= 300 and a < 700:
> >             m = 1
> >             a = a - 300 - m*100
> >         elif a >= 700 and a < 1100:
> >             m=2
> >             a=...
> >         elif a>=1100:
> >             raise ValueErrorr(ZA jumber is not recognized)
> > ```
> 
> It is not really a "general rule" ... It can be used if the nuclide with more than one isomeric state has a mass number greater than 200, as usually happens, but not for Hf177m and Hf177n for example. I'm not sure how to include this nuclide and other similar cases

You are right. If we cannot find a general rule, then let's forget it. I suggest raising a warning when the method is nndc and the mass number is larger th

sandy/zam.py
```diff
@@ -510,6 +524,9 @@ def expand_za(za, method="nndc", meta=0):
         if a >= 300:
             m = 1
             a = a - 300 - m*100
+        while a > 3*z:
+            m += 1
+            a -= 100
```
luca-fiorito-11, line 529: I don't really like this solution,  since, as you said, it is not global. I would just raise a warning unless we have a real equation

luca-fiorito-11 (changes requested):

sandy/sections/mf6.py
```diff
@@ -124,7 +124,7 @@ def read_mf6(tape, mat, mt):
         LAW = T.L2  # Distintion between different distribution function
         ZAP = T.C1  # Product identifier
         LIP = T.L1  # Product isomeric state identifier
-        ZAM = zam.za2zam(ZAP, meta=LIP)
+        ZAM = zam.za2zam(ZAP, meta=LIP, method=False)
```
luca-fiorito-11, line 127: @EnricaBelfiore , In theory this should also work with `method="nndc"`. Why did you change it?

EnricaBelfiore: I checked in the Endf6 manual and the ZAP is defined as 1000 ∗ Z + A. This works also with `method="nndc"` if meta=0, but not with higher isomeric states of the product nucleus

luca-fiorito-11: In this case ZAP is always the ZA of the ground state. The metastate level is defined by LIP.
So, if I remove method does it still work?

EnricaBelfiore: No because the default option is `method="nndc"`. If "nndc" is selected, even if you pass `meta=LIP`, the method `za2zam()` will return the metastate evaluated from the ZA number.  To be more clear: in line 556 of the script sandy.zam.py, we initialize `m=0` and we don't use the `meta` parameter... It could be initialized equal to the parameter `meta` but I lose the meaning of the "nncd" method for me. I left this method as the default option, as it is currently impemented in the `develop` branch

luca-fiorito-11 (approved):

PR #180, opened 2022-04-22
added notebook for CategoryCov

luca-fiorito-11:
Main changes to `CategoryCov`
-------
- several method added to the docstring (not all)
- lowered limit to check symmetry to prevent too many errorrs
- property `std` is now method `get_std`
- classmethod `corr2cov` is now a method
- method `eig` is now called `get_eig`
- method `get_corr` now returns a `CategoryCov` and not a `pd.DataFrame`

Notebooks
----
A notebook has been added to simulate the eigenvalue truncation


luca-fiorito-11:
@AitorBengoechea , ping!

PR #187, opened 2022-04-26
Get_errorr update

AitorBengoechea:
Errorr.py:
Now `get_cov` returns a dictionary with the keys being mf and the values of the dictionary `CategoryCov` object. In case there is only one mf, it returns a `CategoryCov` object directly.
The distinction between `nubar` and `xs` is fundamental, since it helps to recognise errors more easily (in this case, it has helped to recognise errors in `njoy.py`). 
On the other hand, in nucleus with more than one mf, the mf argument allows us to select which mf we will receive as output. 

Njoy.py:
Bug: Previously, if the user entered a mt, that mt was applied to all error output. This generated null files or errors when parsing mt=456 with mf=33 or mt=2 with mf=31, to give an example. Now, if `xs=True` and we input different `mt`, they are parsed and if they are all greater than 450, `3/` is automatically written, so `nubar` and `xs` are parsed in the same input.

### DIFF
--- a/sandy/njoy.py
+++ b/sandy/njoy.py
@@ -758,7 +758,7 @@ def _errorr_input(endfin, pendfin, gendfin, errorrout, mat,
     2 /    
     """
     irelco = 0 if relative is False else 1
-    iread = 1 if mt is not None else 0 
+    iread = 1 if mt is not None and mfcov != 31 and (np.array(mt) < 450).all() else 0
     iwt_ = 1 if spectrum_errorr is not None else iwt_errorr
     ign_ = 1 if ek_errorr is not None else ign_errorr
     text = ["errorr"]
@@ -767,7 +767,7 @@ def _errorr_input(endfin, pendfin, gendfin, errorrout, mat,
     text += [f"{mat:d} {ign_:d} {iwt_:d} {printflag:d} {irelco} /"]
     text += [f"{printflag:d} {temp:.1f} /"]
     text += [f"{iread:d} {mfcov} {irespr:d}/"]
-    if iread == 1:  # only specific mts
+    if iread == 1 and mfcov != 31 and (np.array(mt) < 450).all():  # only specific mts
         mtlist = [mt] if isinstance(mt, int) else mt
         nmt = len(mtlist)
         text += [f"{nmt:d} 0 /"]
@@ -992,7 +992,7 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
     0/
 
     Test keyword `mt` as `list`
-    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=[1, 2]))
+    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=[1, 2], xs=False))
     groupr
     20 21 0 0 /
     22 9237 0 2 0 1 1 0 /
@@ -1005,7 +1005,7 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
     0/       
 
     Test keyword `mt` as `int`
-    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=2))
+    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=2, xs=False))
     groupr
     20 21 0 0 /
     22 9237 0 2 0 1 1 0 /
@@ -1046,6 +1046,9 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
     if mt is None:
         text += ["3/"]  # by default process all cross sections (MF=3)
     else:
+        # Compute mf=33 and mf=31 together
+        if kwargs["xs"] and (np.array(mt) > 450).all():
+            text += ["3/"]
         mtlist = [mt] if isinstance(mt, int) else mt
         for mt_ in mtlist:
             text += [f"3 {mt_:d} /"]
--- a/sandy/errorr.py
+++ b/sandy/errorr.py
@@ -131,21 +131,34 @@ class Errorr(_FormattedFile):
         data = pd.concat(data, axis=1).fillna(0)
         return sandy.Xs(data)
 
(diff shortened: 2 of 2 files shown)


AitorBengoechea:
The program is tested for U-238, which contains mf=31, 33, 34 and 35. In the case of `mf=34` and `mf=35`, a covariance matrix is obtained, but they are not tested (the note in the docstring explains this).

luca-fiorito-11:
> The program is tested for U-238, which contains mf=31, 33, 34 and 35. In the case of `mf=34` and `mf=35`, a covariance matrix is obtained, but they are not tested (the note in the docstring explains this).

can you add this test to the notebooks folder?

AitorBengoechea:
Bug solve: Now when we take all the available mf, never returns a empty list.

AitorBengoechea:
NJOY now processes the `mt` selection as it should, since before, when a `mt` is introduced, it was copied to all the `mf`. Therefore, it could give the option that with mf=31 we were asking NJOY to process mt=2, so the program fails

AitorBengoechea:
I have added an extra condition in `_group_input` and `_errorr_input` so that if the user enters the wrong mt, they are corrected or at least not repeated. (With test include)

sandy/njoy.py
```diff
@@ -1043,14 +1091,27 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
                                           spectrum_groupr[1::2]))
         text += [tab1]
         text += ["/"]
-    if mt is None:
+    if mt is None and kwargs["xs"]:
```
luca-fiorito-11, line 1094: I expected xs to be processed by default.
What if xs is not given?  There is no example about it

luca-fiorito-11 (changes requested):

sandy/core/endf6.py
```diff
@@ -1868,7 +1868,13 @@ def get_errorr(self,
         12 /
         1.00000e-05 3.00000e-02 5.80000e-02 1.40000e-01 2.80000e-01 3.50000e-01 6.25000e-01 4.00000e+00 4.80520e+01 5.53000e+03 8.21000e+05 2.23100e+06 1.00000e+07 /
         3/
+        3 452 'nu' /
```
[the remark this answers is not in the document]
AitorBengoechea: In the method `get_gendf`: Line 2339
Bad mt selection in  `_errorr_input`: line 761
Bad mt selection in  `_groupr_input`: line 1044 

The only missing test  is on `get_errorr`.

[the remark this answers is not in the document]
AitorBengoechea: Line 2021: Keywords `mt` and `groupr` are incompatible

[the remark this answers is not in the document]
AitorBengoechea: The most logical option, in my opinion, is to process all the mt and do the following: a warning saying that the mt will have to be processed by the user or create lines of code at the end that filter the mt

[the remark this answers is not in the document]
luca-fiorito-11: Then let's just make a test showing the incompatibility and let it be for the moment

sandy/njoy.py
```diff
@@ -755,10 +757,30 @@ def _errorr_input(endfin, pendfin, gendfin, errorrout, mat,
     0 293.6 /
     1 33 1/
     1 0 /
-    2 /    
+    2 /
+
+    Test of wrong mt number:
+    >>> with pytest.raises(SyntaxError): sandy.njoy._errorr_input(20, 21, 0, 22, 9237, mt=455)
+    >>> print(sandy.njoy._errorr_input(20, 21, 0, 22, 9237, mt=[102,455]))
```
luca-fiorito-11, line 764: Why don't we get a error/warning here?

luca-fiorito-11: I would just raise a error right away if one of the banned_mt is found in the mt list

sandy/njoy.py
```diff
@@ -1014,7 +1038,30 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
     10000000000.0/
     3 2 /
     0/
-    0/       
+    0/
+
+    Test the wrong mt number:
+    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=[455]))
```
luca-fiorito-11, line 1044: Why not an error?

sandy/njoy.py
```diff
@@ -1014,7 +1038,30 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
...
+    20 21 0 0 /
+    22 9237 0 2 0 1 1 0 /
+    'sandy runs groupr' /
+    293.6/
+    10000000000.0/
+    3/
+    0/
+    0/
+
+    >>> print(sandy.njoy._groupr_input(20, 21, 0, 22, 9237, mt=[102, 455]))
```
luca-fiorito-11, line 1055: Why not an error?

sandy/njoy.py
```diff
@@ -1043,14 +1090,27 @@ def _groupr_input(endfin, pendfin, gendfout, mat,
                                           spectrum_groupr[1::2]))
         text += [tab1]
         text += ["/"]
-    if mt is None:
+    if mt is None and xs:
```
luca-fiorito-11, line 1093: Why not
```
if xs:
   if mt is None:
      ...
   else:
      ...
```

luca-fiorito-11 (changes requested):

sandy/njoy.py
AitorBengoechea: This assumes that the user knows how NJOY is programmed in sandy, which is not always the case. The current programming only raises an error if we are absolutely sure that NJOY is going to fail.

sandy/njoy.py
AitorBengoechea: It has to raise an errorr

sandy/njoy.py
AitorBengoechea: Same as above

PR #189, opened 2022-04-29
Endf get cov

AitorBengoechea:
First step towards the direct extraction of covariance matrices in a single method.

`_get_xs_cov`: Extract the covariance matrix based on the information of mf=33 and mf=31. Tested with `sandy.XsCov`, since the original pipeline was based on this class. 

`_get_lpc_cov`: Extract the covariance matrix based on the information of mf=34. It will go in a separate branch.

`_get_edistr_cov`: Extract the covariance matrix based on the information of mf=35. It will go in a separate branch.

`get_cov`: Get it in a dictionary {mf: `CategoryCov`}. If only one mf exists, it automatically returns the `CategoryCov` object. To test it, tests identical to `_get_xs_cov` were used. Once the branch https://github.com/AitorBengoechea/sandy/tree/errorr_get_cov is in sandy, I will proceed to test nubar and cov with processing.

### DIFF
--- a/sandy/core/endf6.py
+++ b/sandy/core/endf6.py
@@ -2476,11 +2476,88 @@ If you want to process 0K cross sections use `temperature=0.1`.
             }
         return cov_info
 
-    def _get_nubar_cov(self):
-        return
+    def _get_xs_cov(self, mf=[31, 33], mt=range(1, 10000),
+                    mat=range(1, 10000)):
+        """
+        Extract from endf6 file the cross section/nubar covariance matrices.
 
-    def _get_xs_cov(self):
-        return
+        Parameters
+        ----------
+        mf : `int` or `list`, optional
+            MF number. The default is [31, 33].
+        mt : `int` or `list`, optional
+            MT number. The default is range(1, 10000).
+        mat : `int` or `list`, optional
+            MAT number. The default is range(1, 10000).
+
+        Returns
+        -------
+        `sandy.CategoryCov`
+            Covariance matrix.
+
+        Examples
+        --------
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922380)
+        >>> cov = endf6._get_xs_cov()
+        >>> xs_cov = sandy.XsCov.from_endf6(endf6)
+        >>> assert (cov.data == xs_cov).all().all()
+        """
+        listmf_ = [mf] if isinstance(mf, int) else mf
+        listmt_ = [mt] if isinstance(mt, int) else mt
+        listmat_ = [mat] if isinstance(mat, int) else mat
+        tape = self.filter_by(listmf=listmf_, listmt=listmt_, listmat=listmat_)
+        data = []
+        for mat, mf, mt in tape.data:
+            sec = tape.read_section(mat, mf, mt)
+            for sub in sec["SUB"].values():
+                mat1 = sub['MAT1'] if sub['MAT1'] != 0 else mat
+                mt1 = sub['MT1']
+                covs = []
+                # Loop NI-type covariances
+                for nisec in sub["NI"].values():
+                    lb = nisec["LB"]
+                    if lb == 5:
+                        if nisec["LS"] == 0:
+                            foo = sandy.EnergyCov.from_lb5_asym
+                        else:
+                            foo = sandy.EnergyCov.from_lb5_sym
+                        cov = foo(nisec["EK"], nisec["FKK"])
+                    elif lb == 1:
+                        foo = sandy.EnergyCov.from_lb1
+                        cov = foo(nisec["EK"], nisec["FK"])
+                    elif lb == 2:
+                        foo = sandy.EnergyCov.from_lb2
+                        cov = foo(nisec["EK"], nisec["FK"])
+                    elif lb == 6:
+                        foo = sandy.EnergyCov.from_lb6
+                        cov = foo(nisec["EK"], nisec["EL"], nisec["FKL"])
+                    else:
+                        logging.warning(f"skip 'LB={lb}' covariance for"
+                                        f" [({mat}/{mt}), ({mat1}/{mt1})]")
+                        continue
+                    covs.append(cov)
+                if not covs:
+                    continue
+                cov = sandy.EnergyCov.sum_covs(*covs)
+                idx = pd.MultiIndex.from_product(
+                    [[mat], [mt], list(cov.data.index)],
+                    names=["MAT", "MT", "E"],
+                    )
+                idx1 = pd.MultiIndex.from_product(
+                    [[mat], [mt1], list(cov.data.columns)],
+                    names=["MAT1", "MT1", "E1"],
+                    )
+                df = pd.DataFrame(cov.data.values, index=idx, columns=idx1) \
+                       .stack(level=["MAT1", "MT1", "E1"]) \
+                       .rename("VAL") \
+                       .reset_index()
+                data.append(df)
+        if not data:
+            return None
+        df = pd.concat(data)
+        return sandy.CategoryCov.from_stack(df, index=["MAT", "MT", "E"],
+                                            columns=["MAT1", "MT1", "E1"],
+                                            values='VAL')
 
     def _get_lpc_cov(self):
         return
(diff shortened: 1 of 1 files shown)


AitorBengoechea:
As the functions `_get_lpc_cov` and `_get_edistr_cov` are created and the branch https://github.com/AitorBengoechea/sandy/tree/errorr_get_cov accepted, more tests will be carried out for different situations.

AitorBengoechea:
Priorities for new sampling: 

1) https://github.com/AitorBengoechea/sandy/tree/errorr_get_cov

2) https://github.com/AitorBengoechea/sandy/tree/endf_get_cov

3) I will upload the script where I have stored the new `sampling.py`.

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        else:
+            listmf_ = list(self.to_series().index.get_level_values("MF")
+                               .intersection([31, 33, 34, 35]))
+        list_process = [process_mf] if isinstance(process_mf, int) else process_mf
+        # Dividing the list information:
+        listmf_ = list(pd.Index(listmf_).difference(pd.Index(list_process)))
+        if len(list_process) != 0:
+            kwds_njoy["nubar"] = True if 31 in list_process else False
+            kwds_njoy["xs"] = True if 33 in list_process else False
+            kwds_njoy["chi"] = True if 34 in list_process else False
```
luca-fiorito-11, line 2619: Mubar is 34, chi is 35

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+            kwds_njoy["mubar"] = True if 35 in list_process else False
```
luca-fiorito-11, line 2620: We don't need to run errorr for mubar and chi. We will always read them directly from the endf6 file

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+            out = self.get_errorr(**kwds_njoy)
+            cov = out.get_cov(mf=list_process)
```
luca-fiorito-11, line 2622: Shouldn't mt be an option in `get_cov`? It is in `_get_xs_cov`

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        if len(listmf_) != 0:
+            cov = {} if len(list_process) == 0 else cov
+            if 31 in listmf_:
+                cov[31] = self._get_xs_cov(mf=[31])
+            if 33 in listmf_:
+                cov[33] = self._get_xs_cov(mf=[33])
```
luca-fiorito-11, line 2628: By default mf31 and 33 should be taken from errorr

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+            If only one mf is introduced, the method returns the
+            `sandy.CategoryCov` object directly.
+
+        Examples
+        --------
+        Nubar covariance matrix from endf6 file:
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922380)
+        >>> xs_cov = sandy.XsCov.from_endf6(endf6)
+        >>> nubar_cov = endf6.get_cov(mf=31)
+        >>> assert (nubar_cov.data.values == xs_cov.loc[(9237, 456), (9237, 456)].values).all().all()
```
luca-fiorito-11, line 2595: I don't like testing against deprecated functions

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+
+        Xs covariance matrix from endf6 file:
+        >>> endf_xs_cov = endf6.get_cov(mf=33)
+        >>> size = len(endf_xs_cov.data)
+        >>> assert (endf_xs_cov.data.values == xs_cov.iloc[:size, :size].values).all().all()
+
+        Both together:
+        >>> endf_xs_cov = endf6.get_cov(mf=[31, 33])
```
luca-fiorito-11, line 2603: This tests imply that the default is to read mf31 and mf33 from the endf6 file, which shouldn't be

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        Returns
+        -------
+        `sandy.CategoryCov`
+            Covariance matrix.
+
+        Examples
+        --------
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922380)
+        >>> cov = endf6._get_xs_cov()
+        >>> xs_cov = sandy.XsCov.from_endf6(endf6)
```
luca-fiorito-11, line 2502: Don't test against deprecated functions

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        >>> assert (cov.data == xs_cov).all().all()
+        """
+        listmf_ = [mf] if isinstance(mf, int) else mf
+        listmt_ = [mt] if isinstance(mt, int) else mt
+        listmat_ = [mat] if isinstance(mat, int) else mat
+        tape = self.filter_by(listmf=listmf_, listmt=listmt_, listmat=listmat_)
+        data = []
+        for mat, mf, mt in tape.data:
+            sec = tape.read_section(mat, mf, mt)
+            for sub in sec["SUB"].values():
```
luca-fiorito-11, line 2512: If I give in input `mf=5`, this will fail

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        ----------
+        mf : `int` or `list`, optional
+            MF number. The default is [31, 33].
+        mt : `int` or `list`, optional
+            MT number. The default is range(1, 10000).
+        mat : `int` or `list`, optional
+            MAT number. The default is range(1, 10000).
+
+        Returns
+        -------
```
luca-fiorito-11, line 2494: You can also get None

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
             'chi': run_chi if chi else False,
             }
         return cov_info
+
+    def _get_xs_cov(self, mf=[31, 33], mt=range(1, 10000),
```
luca-fiorito-11, line 2479: Why do we need a mf? Thus could be by default 33. And then have another method `get_nubar_xs` with 31 as default. In theory these two methods should both call a common third method

sandy/core/endf6.py
```diff
@@ -2475,3 +2475,162 @@ def covariance_info(self, nubar=True, xs=True, mubar=True, chi=True):
...
+        -------
+        `sandy.CategoryCov`
+            Covariance matrix.
+
+        Examples
+        --------
+        >>> endf6 = sandy.get_endf6_file('jeff_33','xs', 922380)
+        >>> cov = endf6._get_xs_cov()
+        >>> xs_cov = sandy.XsCov.from_endf6(endf6)
+        >>> assert (cov.data == xs_cov).all().all()
```
luca-fiorito-11, line 2503: More test needed...at least to check the input options

luca-fiorito-11 (changes requested):

sandy/core/endf6.py
AitorBengoechea: That's right, I need to update the docstring.

sandy/core/endf6.py
AitorBengoechea: At the moment this branch is not able to do that. For that you need the updates of the branch https://github.com/AitorBengoechea/sandy/tree/errorr_get_cov

sandy/core/endf6.py
AitorBengoechea: I have tested it this way to ensure that we get the same information as in the original pipeline. This test will disappear in the future, but for the moment it seemed to me the most reliable way to test the whole matrix.

sandy/core/endf6.py
AitorBengoechea: I think we have to trust that the user knows what he or she is doing. In NJOY, if you produce an incorrect input (for example, in ERRORR with mf=31 asking it to parse mt=2), NJOY fails and it doesn't mean that it is badly programmed, but that we as a user are asking it to do something that doesn't make sense. 

However, if you want before processing, I can set an option to return nothing and print a message.

sandy/core/endf6.py
AitorBengoechea: As the other branches are introduced, the tests will be updated.

AitorBengoechea:
With the new pull request the order of priorities is as follows:
Priorities for new sampling: 

1) https://github.com/AitorBengoechea/sandy/tree/errorr_get_cov

2) https://github.com/AitorBengoechea/sandy/tree/get_lpc_cov

3) https://github.com/AitorBengoechea/sandy/tree/get_edistr_cov

4) https://github.com/AitorBengoechea/sandy/tree/endf_get_cov

5) I will upload the script where I have stored the new `sampling.py`.This script will create the samples without disturbing them. 

6) For the perturbation in the sampling, the need to accept the https://github.com/AitorBengoechea/sandy/tree/fy branch

7) Review of the `custom_perturbation` of the other classes

8) Parallelising and perturbing samples

PR #195, opened 2022-05-07
Pipeline changes

EnricaBelfiore:
Pdf option added 
---

The option to select the distribution for the sampling procedure is added to the pipeline. The perturbation coefficients are now extracted with the method `sandy.CategoryCov.sampling()` according to the chosen distribution and covariance matrix.

### DIFF
--- a/sandy/sampling.py
+++ b/sandy/sampling.py
@@ -604,11 +604,6 @@ def parse(iargs=None):
                         default='normal',
                         help="draw samples according to the chosen distribution "
                              "(default = 'normal')")
-    parser.add_argument('--relative',
-                        type=bool,
-                        default=True,
-                        help="flag to switch between relative and absolute covariance matrix "
-                             "(default = True)")
     parser.add_argument('--njoy',
                         type=lambda x: is_valid_file(parser, x),
                         default=None,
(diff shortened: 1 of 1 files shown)


sandy/sampling.py
```diff
@@ -752,7 +757,10 @@ def extract_samples(ftape, covtape):
         covtape = covtape.filter_by(listmat=init.mat, listmf=[1,33], listmt=listmt)
         xscov = XsCov(covtape.get_cov(multigroup=False).data) if isinstance(covtape, sandy.errorr.Errorr) else XsCov.from_endf6(covtape)
         if not xscov.empty:
-            PertXs = xscov.get_samples(init.samples, eig=init.eig, seed=init.seed33)
+            PertXs = sandy.CategoryCov(xscov).sampling(init.samples, tolerance=0, seed=init.seed33, pdf=init.pdf).data.T
+            PertXs.rename(columns=dict(zip(range(0, init.samples), range(1, init.samples + 1))), inplace=True)
```
luca-fiorito-11, line 761: Why do you rename columns here and then again in line 762?

sandy/sampling.py
```diff
@@ -808,9 +816,13 @@ def sampling(iargs=None):
 #            ftape = ftape.delete_sections((None, 3, None)). \
 #                          add_sections(ptape.filter_by(listmf=[3])). \
 #                          add_sections(ptape.filter_by(listmf=[1], listmt=[451]))
-        pxs = xscov.get_samples(init.samples, eig=init.eig, seed=init.seed33)
-        cn = sandy.Samples(pxs).condition_number
+        pxs = xscov.sampling(init.samples, pdf=init.pdf, seed=init.seed33, tolerance=0)
+        cn = pxs.condition_number
         print(f"Condition number : {cn:>15}")
+        pxs = pxs.data.T
+        pxs.rename(columns=dict(zip(range(0, init.samples), range(1, init.samples + 1))), inplace=True)
```
luca-fiorito-11, line 823: Same comment as above. What does this line do? Is there a simpler way to do it?

sandy/sampling.py
```diff
@@ -599,6 +599,11 @@ def parse(iargs=None):
                         metavar="{1,..,999}",
                         help="draw samples only from the selected MT sections "
                              "(default = keep all)")
+    parser.add_argument('--pdf',
+                        type=str,
+                        default='normal',
+                        help="draw samples according to the chosen distribution "
```
luca-fiorito-11, line 605: Mention what options you have.
In theory you should also check that only one of the selected options is chosen. And ot should be co verteld to lowercase.

The parser allows you to do all this. Check the manual

luca-fiorito-11 (changes requested):

luca-fiorito-11 (approved):

PR #199, opened 2022-05-12
Globalization of gls updates

AitorBengoechea:
The aim of this branch is to make the GLS that sandy has fully global (also, for constrained gls update). To do this, the `gls_update` methods have been updated to work with sensitivity vectors. @EnricaBelfiore will implement the changes and test the functions with sensitivity vectors.


AitorBengoechea:
Future work: Once the implementation is done, test the fy.py methods and update if it is necessary

EnricaBelfiore:
Some comments on the last commit:

- With the new implementation of the GLS update, the design matrix S can be MXN with M that could be also equals to 1. I specified it in the docstring but maybe it is redundant. What do you think?
- The `sandy.constrained_gls_update()` function and all the private functions to perform it are removed. This is because, in terms of implementation, the GLS update formula and constrained GLS update formula are equivalent if Vy_extra is zero (default is `None`). 
- I inserted a warning in the `sandy.gls._gls_G_inv()` function. This is because when the determinant of the matrix `S*V_prior*S^T + Vy_extra` is close to zero (e.x. 1e-19) the method `np.linalg.inv()` will not work properly. Serching on internet I found that in this case it is possible to perform the pseudo-inverse of a matrix with the method `np.linalg.pinv()` ([[inverse matrix np.linalg.inv in Python, gives weird results, why?](https://stackoverflow.com/questions/66094987/inverse-matrix-np-linalg-inv-in-python-gives-weird-results-why)](https://stackoverflow.com/questions/66094987/inverse-matrix-np-linalg-inv-in-python-gives-weird-results-why)). Is it more convinient to use the latter in that case?
- The tests in the `sandy.fy.py` script are wrong. I kept them as they are but they will be fixed in another pull request.

luca-fiorito-11 (changes requested):
Just minor changes

sandy/fy.py
```diff
@@ -895,7 +895,11 @@ def ishikawa_factor(self, zam, e, Vy_extra,
             model_sensitivity_object = decay_data
         S = _gls_setup(model_sensitivity_object, kind)
         # Perform Ishikawa factor:
-        ishikawa = sandy.ishikawa_factor(S, Vx_prior, Vy_extra)
+        Vy_extra_ = pd.DataFrame(Vy_extra)
+        index = Vy_extra_.index
+        S = S.reindex(index=index, columns=Vx_prior.index).fillna(0)
+        ishikawa = sandy.ishikawa_factor(S.values, Vx_prior.values, Vy_extra_.values)
```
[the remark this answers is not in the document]
EnricaBelfiore: This will be done in another PR

sandy/gls.py
```diff
@@ -22,7 +21,7 @@
         ]
 
 x_prior = [1, 2, 3]
-y_extra = pd.Series([2, 3, 4], index=[1, 2, 3])
+y_extra = np.array([2, 3, 4])
```
luca-fiorito-11, line 24: @EnricaBelfiore, what are all these things? If not needed we can remove them

sandy/core/cov.py
```diff
@@ -1962,6 +1958,73 @@ def sparse_tables_inv(a, rows=1000):
     return invert_matrix
 
 
+def segmented_pivot_table(data_stack, index, columns, values, rows=10000000):
```
luca-fiorito-11, line 1961: @EnricaBelfiore , do we use this?

sandy/core/cov.py
```diff
@@ -1885,7 +1881,7 @@ def sparse_tables_dot(a, b, rows=1000):
     return dot_product
 
 
-def sparse_tables_dot_multiple(matrix_list, rows=1000):
+#def sparse_tables_dot_multiple(matrix_list, rows=1000):
```
luca-fiorito-11, line 1884: @EnricaBelfiore , why commenting it and not removing it?

luca-fiorito-11 (approved):

sandy/gls.py
EnricaBelfiore: They have been inserted and used in the docstring example of `sandy.gls._y_calc()`. They will be removed (and the function as well) in another PR

sandy/core/cov.py
EnricaBelfiore: Yes, it is used in the class method `sandy.CategoryCov.from_stack()`

sandy/core/cov.py
EnricaBelfiore: My fault, I forgot to delete the comments but the function is no more used and it will be removed

luca-fiorito-11 (approved):

PR #202, opened 2022-05-18
Custom perturbation with dataframes

AitorBengoechea:
@luca-fiorito-11:
Now custom perturbation is coherent with njoy results and can perturb all the xs.

### DIFF
--- a/sandy/core/xs.py
+++ b/sandy/core/xs.py
@@ -156,10 +156,225 @@ class Xs():
         df = pd.DataFrame(xsnew, index=enew, columns=df.columns)
         return self.__class__(df)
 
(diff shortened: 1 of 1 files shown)


AitorBengoechea:
Now we have to decide if we have to erase the function of `_custom_single_perturbation`

AitorBengoechea:
Thes test that are falling are replace by the new custom perturbation (with dataframes in the other branch)

AitorBengoechea:
Thes test that are falling are replace by the new custom perturbation (with dataframes in the other branch)

AitorBengoechea:
When energy distribution custom perturbation is add, all the test will be passed

PR #206, opened 2022-05-19
Lpc perturbation

AitorBengoechea:


AitorBengoechea:
The test that falls is solve in another branch (edistr_perturbation branch)

PR #210, opened 2022-05-24
bug in reading metastable state of FY

EnricaBelfiore:
FY from_endf6 and to_endf6
---

-  Added metastable state in the evaluation of the ZAM number of each fissioning nuclide in the method `sandy.Fy.from_endf6().`

- Correction of the method `sandy.Fy.to_endf6()` to not overwrite the `sandy.Endf6` input instance.

### DIFF
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -1012,19 +1012,19 @@ class Fy():
         >>> tape = sandy.get_endf6_file("jeff_33", "nfpy", 'all')
         >>> fy = sandy.Fy.from_endf6(tape)
         >>> fy.data.query("ZAM==952421 & MT==454 & E==0.0253").head()
-                MAT   MT         ZAM    ZAP           E          FY         DFY
-        56250  9547  454 9.52421e+05  10010 2.53000e-02 3.32190e-05 1.17790e-05
-        56251  9547  454 9.52421e+05  10020 2.53000e-02 1.01520e-05 3.52080e-06
-        56252  9547  454 9.52421e+05  10030 2.53000e-02 1.60000e-04 5.00220e-05
-        56253  9547  454 9.52421e+05  20030 2.53000e-02 0.00000e+00 0.00000e+00
-        56254  9547  454 9.52421e+05  20040 2.53000e-02 2.10000e-03 6.64080e-04
+                MAT   MT     ZAM    ZAP           E          FY         DFY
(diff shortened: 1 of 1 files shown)


sandy/fy.py
```diff
@@ -1006,26 +1006,41 @@ def from_endf6(cls, endf6, verbose=False):
...
+            else:
+                if verbose:
+                    logging.info(f"reading 'MAT={mat}/MT={mt}'...")
+                for e in sec["E"]:
+                    for zap in sec["E"][e]["ZAP"]:
+                        fy = sec["E"][e]["ZAP"][zap]["FY"]
+                        dfy = sec["E"][e]["ZAP"][zap]["DFY"]
+                        values = (mat, mt, zap, e, fy, dfy)
+                        data.append(dict(zip(["MAT", "MT", "ZAP", "E", "FY", "DFY"], values)))
+        df_zam = pd.DataFrame({"MAT": endf6.mat, "ZAM": zam})
```
luca-fiorito-11, line 1037: here you are assuming that `endf6.mat` is sorted in the same order as the for loop `endf6.keys`.
Is it true?

I guess it makes more sense if zam is a dictionary with mat as a keys.
Then your `df_zam` should be something like `pd.DataFrame(zam, names=["MAT", "ZAM"])`.
Even better if instead of zam you call it dict_zam

sandy/fy.py
```diff
@@ -1006,26 +1006,41 @@ def from_endf6(cls, endf6, verbose=False):
...
         .. note:: Both independent and cumulative fission product yields are
                   loaded, if found.
+
+        Examples
+        --------
+        >>> tape = sandy.get_endf6_file("jeff_33", "nfpy", 'all')
+        >>> fy = sandy.Fy.from_endf6(tape)
+        >>> fy.data.query("ZAM==952421 & MT==454 & E==0.0253").head()
+                MAT   MT         ZAM    ZAP           E          FY         DFY
+        56250  9547  454 9.52421e+05  10010 2.53000e-02 3.32190e-05 1.17790e-05
```
luca-fiorito-11, line 1016: ZAM must be a `int`

luca-fiorito-11 (changes requested):

sandy/fy.py
EnricaBelfiore: Yes, it is true but I inserted the dictionary in the last commit

luca-fiorito-11 (approved):

PR #211, opened 2022-05-29
Dot product

EnricaBelfiore:
`_y_calc_` replaced with dot product
---

- Removed `sandy.gls._y_calc()` function 
- Fixed docstrings mismatch between `sandy.Fy.get_mass_yield()` and `sandy.Fy.get_chain_yield()`

### DIFF
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -328,8 +328,8 @@ class Fy():
         >>> zam = [591480, 591481, 601480]
         >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
         >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
-        >>> npfy = sandy.Fy(minimal_fytest_2)
-        >>> chain = npfy.get_chain_yield(942390, 500e3, decay_fytest).iloc[0].round(2) 
+        >>> nfpy = sandy.Fy(minimal_fytest_2)
+        >>> chain = nfpy.get_chain_yield(942390, 500e3, decay_fytest).iloc[0].round(2) 
         >>> assert chain == 0.6
         """
         # Filter FY data:
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -290,20 +290,18 @@ class Fy():
 
         Examples
         --------
-        >>> zam = [591480, 591481, 601480]
-        >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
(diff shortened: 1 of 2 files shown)


sandy/fy.py
```diff
@@ -290,20 +290,18 @@ def get_mass_yield(self, zam, e):
 
         Examples
         --------
-        >>> zam = [591480, 591481, 601480]
-        >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
-        >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
-        >>> npfy = sandy.Fy(minimal_fytest_2)
-        >>> chain = npfy.get_chain_yield(942390, 500e3, decay_fytest).iloc[0].round(2) 
-        >>> assert chain == 0.6
+        >>> tape_nfpy = sandy.get_endf6_file("jeff_33",'nfpy','all')
```
luca-fiorito-11, line 293: @EnricaBelfiore , why calling all fys where we are only interested in 922350?

luca-fiorito-11 (changes requested):

sandy/fy.py
```diff
@@ -327,17 +325,20 @@ def get_chain_yield(self, zam, e, decay_data, **kwargs):
...
-        >>> tape_nfpy = sandy.get_endf6_file("jeff_33",'nfpy','all')
-        >>> nfpy = Fy.from_endf6(tape_nfpy)
-        >>> nfpy.get_mass_yield(922350, 0.0253).loc[148]
-        0.0169029147
+        >>> zam = [591480, 591481, 601480]
+        >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
+        >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
+        >>> nfpy = sandy.Fy(minimal_fytest_2)
+        >>> chain = nfpy.get_chain_yield(942390, 500e3, decay_fytest).iloc[0].round(2) 
+        >>> assert chain == 0.6
```
luca-fiorito-11, line 333: @EnricaBelfiore , I don't like much this test. With a iloc I don't even know what I am looking at. From this test I cannot understand if the methodis correct and it can be removed. It would be nice to have something similar to the test in get_mass_yield to show the difference. Later for another PR.

luca-fiorito-11 (approved):

sandy/fy.py
EnricaBelfiore: This test and many others you commented on come from the previous implmentations. I will modify them in future PRs and check for other inconsistencies.

PR #212, opened 2022-05-29
Bug in apply_qmatrix and apply_bmatrix

EnricaBelfiore:
Bug in `sandy.Fy.apply_bmatrix()` and `sandy.Fy.apply_qmatrix()`
---
The mentioned methods used uncertainties as variances  rather than standard deviations to build the covariance matrix. This problem is solved in this PR and some docstrings are adjusted.

### DIFF
--- a/sandy/fy.py
+++ b/sandy/fy.py
@@ -325,12 +325,13 @@ class Fy():
 
         Examples
         --------
-        >>> zam = [591480, 591481, 601480]
+        >>> zam = [591480, 591481, 601480, 561480, 571480, 571490, 581480]
         >>> decay_minimal = sandy.get_endf6_file("jeff_33", 'decay', zam)
         >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
-        >>> nfpy = sandy.Fy(minimal_fytest_2)
-        >>> chain = nfpy.get_chain_yield(942390, 500e3, decay_fytest).iloc[0].round(2) 
-        >>> assert chain == 0.6
+        >>> tape_nfpy = sandy.get_endf6_file("jeff_33", 'nfpy', 922350)
+        >>> nfpy = Fy.from_endf6(tape_nfpy)
+        >>> nfpy.get_chain_yield(922350, 0.0253, decay_fytest).loc[148]
+        0.01692277272
         """
         # Filter FY data:
         conditions = {'ZAM': zam, "E": e, 'MT': 454}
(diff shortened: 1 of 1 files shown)


sandy/fy.py
```diff
@@ -441,20 +442,20 @@ def apply_bmatrix(self, zam, e, decay_data, keep_fy_index=False):
             Energy to which calculations are to be applied.
         decay_data : `sandy.DecayData`
             Radioactive nuclide data for several isotopes.
-        keep_fy_index=False : `bool`, optional
+        keep_fy_index: `bool`, optional, default is `False`
```
luca-fiorito-11, line 445: It is weird to call it `keep_fy_index` since the fy dataframe does not have a real index. To be changed in a new PR

sandy/fy.py
```diff
@@ -463,23 +464,23 @@ def apply_bmatrix(self, zam, e, decay_data, keep_fy_index=False):
         >>> decay_fytest = sandy.DecayData.from_endf6(decay_minimal)
         >>> npfy = Fy(minimal_fytest_2)
         >>> npfy_pert = npfy.apply_bmatrix(942390, 5.00000e+05, decay_fytest)
-        >>> npfy_pert.data[npfy_pert.data.MT == 454]
-        	 MAT	 MT	   ZAM	   ZAP	          E	         FY	        DFY
-        3	9437	454	942390	591480	5.00000e+05	8.00000e-01	4.00000e-02
-        4	9437	454	942390	591481	5.00000e+05	1.00000e+00	5.00000e-02
-        5	9437	454	942390	601480	5.00000e+05	-1.60000e+00	1.00000e-01
-        6	9437	454	942390	621480	5.00000e+05	-2.00000e-01	1.00000e-02
+        >>> npfy_pert.data.query("MT==454")
```
luca-fiorito-11, line 467: We might be using all fys instead of a `minimal_fytest_2`.

sandy/fy.py
```diff
@@ -504,10 +505,10 @@ def apply_bmatrix(self, zam, e, decay_data, keep_fy_index=False):
         data = data.loc[~mask]
         fy_data = fy_data.reindex(columns).fillna(0)
         cov_data = cov_data.reindex(columns).fillna(0)
-        cov_data = sandy.CategoryCov.from_var(cov_data)
+        cov_data = sandy.CategoryCov.from_stdev(cov_data)
```
luca-fiorito-11, line 508: This is a covariance matrix. I would change the variable name to show it is not what cov_data was before.

luca-fiorito-11 (approved):

PR #214, opened 2022-06-07
Process XS with only MF = 32 section

AitorBengoechea:
As the test take a lot of time, I upload a test showing the proccess


luca-fiorito-11:
outdated

PR #225, opened 2022-11-04
Handle "gamma dose" in ALEPH-2.9 outputs

luca-fiorito-11:
the gamma dose output lines are removed from the output text when parsing tables.


luca-fiorito-11:
closed because merged to wrong branch

PR #230, opened 2022-11-18
Avoid negative samples in Uniform distribution

EnricaBelfiore:
In this pull request the `cutzeros` option is inserted to avoid the symmetrical cut of the uniform distribution. To do so, the standard deviations larger than  1/a are set equal to 1/a, where - a is the lower bound of the distribution y

### DIFF
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -633,7 +633,7 @@ class CategoryCov():
         return np.log(mu_**2 / np.sqrt(np.diag(self.data) + mu_**2))
 
     def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
-                 tolerance=None, relative=True, truncate=True):
+                 tolerance=None, relative=True, truncate=True, cutzeros=False):
         """
         Extract perturbation coefficients according to chosen distribution with
         covariance from given covariance matrix. See note for non-normal
@@ -667,6 +667,10 @@ class CategoryCov():
             flag to perform a symmetrical cut. All the negative perturbation 
             coefficients will be set equal to 0 and the ones larger than 2 will
             be set equal to 2
+        cutzeros : `bool`, optional, default is `False`
+            flag to adjust the distrtibution standard deviation in such a way
+            that negative samples are avoited. This option is only available
+            for `pdf=='uniform'`
 
         Returns
         -------
@@ -790,6 +794,13 @@ class CategoryCov():
         0 2.00000e+00 2.00000e+00
         1 7.13927e-01 1.07147e+00
         2 5.15435e-01 1.64683e+00
+
+        `cutzeros` kwarg usage:
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, pdf='uniform', truncate=False, cutzeros=True)
+                    0           1
+        0 3.60539e-01 1.44987e+00
+        1 3.89505e-02 8.40407e-01
+        2 9.26437e-01 9.70854e-01
         """
         dim = self.data.shape[0]
         pdf_ = pdf if pdf != 'lognormal' else 'normal'
@@ -800,6 +811,11 @@ class CategoryCov():
         # pdf
         if pdf == 'uniform':
             to_decompose = self.__class__(np.diag(np.diag(self.data)))
+            if cutzeros:
+                a = np.sqrt(12) / 2
+                std = pd.DataFrame(np.sqrt(np.diag(np.diag(self.data))))
+                condition = std < 1 / a
+                to_decompose = self.__class__(to_decompose.data.where(condition, (1 / a**2)))
         elif pdf == 'lognormal':
             ones = np.ones(self.data.shape[0])
             to_decompose = self.log2norm_cov(ones)
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -687,17 +687,17 @@ class CategoryCov():
         Examples
         --------
         Draw 3 sets of samples using custom seed:
-        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, relative=False)
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, relative=False).data + 1
                      0            1
-        0 -1.74945e+00 -3.13159e+00
-        1  2.86073e-01  1.06836e-01
-        2  4.84565e-01 -9.91209e-02
+        0 -7.49455e-01 -2.13159e+00
+        1  1.28607e+00  1.10684e+00
+        2  1.48457e+00  9.00879e-01
 
-        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, rows=1, relative=False)
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, rows=1, relative=False).data + 1
                      0            1
-        0 -1.74945e+00 -3.13159e+00
-        1  2.86073e-01  1.06836e-01
-        2  4.84565e-01 -9.91209e-02
+        0 -7.49455e-01 -2.13159e+00
+        1  1.28607e+00  1.10684e+00
+        2  1.48457e+00  9.00879e-01
 
         >>> sample = sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(1000000, seed=11, relative=False)
         >>> sample.data.cov()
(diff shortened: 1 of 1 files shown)


sandy/core/cov.py
```diff
@@ -800,6 +811,11 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
         # pdf
         if pdf == 'uniform':
             to_decompose = self.__class__(np.diag(np.diag(self.data)))
+            if cutzeros:
+                a = np.sqrt(12) / 2
```
luca-fiorito-11, line 815: please add a comment to say what this means

sandy/core/cov.py
```diff
@@ -800,6 +811,11 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
         # pdf
         if pdf == 'uniform':
             to_decompose = self.__class__(np.diag(np.diag(self.data)))
+            if cutzeros:
+                a = np.sqrt(12) / 2
+                std = pd.DataFrame(np.sqrt(np.diag(np.diag(self.data))))
```
luca-fiorito-11, line 816: minor comment, but i don't like calling this `std` since technically it is a matrix with the standard deviations on the diagonal.

Second comment: I would do
```
std = pd.DataFrame(np.sqrt(np.diag(np.diag(to_decompose.data))))
```

Third comment: independently from the second comment, since this matrix only contains non-zero values on the diagonal, isn't it an overkill to look for a condition over all elements of the matrix? if it is for simplicity of implementation I would understand.
But, wouldn't it be simpler to do:
```
std = np.sqrt(np.diag(to_decompose.data))
std_modified = np.where(std < a,  std, a)
cov = np.diag(std.modified)
to_decompose = self.__class__(cov, index=to_decompose.data.index, columns=to_decompose.datacolumns)
```
Maybe not, since at the end you need to reinitialize the dataframe with index and columns.

Up to you.
Let me know

sandy/core/cov.py
```diff
@@ -667,6 +667,10 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
             flag to perform a symmetrical cut. All the negative perturbation 
             coefficients will be set equal to 0 and the ones larger than 2 will
             be set equal to 2
+        cutzeros : `bool`, optional, default is `False`
+            flag to adjust the distrtibution standard deviation in such a way
+            that negative samples are avoited. This option is only available
```
luca-fiorito-11, line 672: typo

sandy/core/cov.py
```diff
@@ -633,7 +633,7 @@ def log2norm_mean(self, mu):
         return np.log(mu_**2 / np.sqrt(np.diag(self.data) + mu_**2))
 
     def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
-                 tolerance=None, relative=True, truncate=True):
+                 tolerance=None, relative=True, truncate=True, cutzeros=False):
```
luca-fiorito-11, line 636: I am not even sure whether to keep it as a keyword argument.
I guess it should just be the default.
@GrimFe , your opinion?

Then, the `truncate` condition should not apply anymore to the uniform distribution, and since it did not apply either for the lognormal distribution, it only remains for the normal one.
For simplicity we might even remove it, since also for the normal distribution we always truncate .

luca-fiorito-11 (changes requested):

luca-fiorito-11:
@GrimFe , can you have a look at this WR?

sandy/core/cov.py
GrimFe: I see your point, but I like leaving the option to the user.
If the point is the number of keyword arguments, then one could think of using only the `truncate` condition, just differently applied to the gaussian distribution (symmetrical cut + setting of the tails to zero or to 2*mean), lognormal distribution (ignored) and uniform distribution (implementing the procedure now conditioned by `cutzeros`).
I agree that by default SANDY should not produce negative samples, but this is a matter of approach.

luca-fiorito-11:
@EnricaBelfiore , there were some problems with github actions that made the check failed.
Now it should be fixed.

sandy/core/cov.py
luca-fiorito-11: Fixed

sandy/core/cov.py
luca-fiorito-11: Then, I guess we can converge on the following:

- all truncations should be applied by default and independently on the distribution
- there is no need for two kwargs `truncate` and `cutzeros`. If kept, they should be merged into one because they are basically doing the same
- keeping them (as @GrimFe prefers) or removing them (as I prefer) will be up to @EnricaBelfiore , making an assessment on the simplicity of consistently merging the two keywords into one, and on the maintainability of the method.

EnricaBelfiore: The second option would require to remove the `relative` kwarg, due to the fact that negative samples are allowed if the sampling is performed with an absolute covariance matrix. If we are ok with removing that, I will remove these kwargs (`relative`, `cutzeros` and `truncate`) and the sampling method will only give positive/relative values

luca-fiorito-11: Relative in theory is needed for sampling MF35 (covariance matrices are in absolute values).
Can't we just truncate between -1 and 1 (instead of 0 and 2) and then recenter into one if relative?

luca-fiorito-11 (changes requested):

sandy/core/cov.py
EnricaBelfiore: In absolute terms I think we should cut between -mean and +mean (of the evaluation and not of the perturbation coefficient distribution) to avoid negative samples... so I would sample only positive values when the covariance is relative and allow negative samples (without symmetrical cut) when the covariance is absolute, removing the `truncate` and `cutzeros` kwargs

luca-fiorito-11: Agreed.
If the covariance is in absolute terms we cannot say a priori if the samples are positive or negative.
So we truncate `if relative==True`.

sandy/core/cov.py
```diff
@@ -831,11 +822,11 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
             samples = np.exp(samples.add(self.log2norm_mean(ones), axis=0))
         elif relative:
             samples += 1
-        if truncate_:
-            lower_bound = samples > 0
-            upper_bound = samples < 2
-            samples = samples.where(lower_bound, 0)
-            samples = samples.where(upper_bound, 2)
+            if not already_cut:
```
luca-fiorito-11: Do we need `already_cut`? 
In principle you can always apply this truncation, independently whether `already_cut=True` or `already_cut=False`. The results should be the same.
I am not against it, but if it useless (or if it saves only a marginal amount of computational time) then I would remove it to simplify the code.

sandy/core/cov.py
```diff
@@ -682,59 +674,63 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
...
-            Therefore, `relative` parameter does not apply to it. The lognormal
-            distribution does not have negative values, so the `truncate`
-            parameter does not apply to it.
+            Therefore, `relative` parameter does not apply to it.
 
         Examples
         --------
         Draw 3 sets of samples using custom seed:
-        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, truncate=False)
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, relative=False)
```
luca-fiorito-11: Well-established test outputs should not vary, otherwise we lose the purpose of the tests.
If the output of the test changes at every commit it doesn't tell us if we are right or not.
In this case you can restore the original output  by doing:
```
sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, relative=False)  - 1
```
Since the only difference is that you re-centered the distribution in zero (to preserve the negative samples)

sandy/core/cov.py
```diff
@@ -776,46 +772,41 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
...
-        Cut negative perturbation coefficients:
-        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, pdf='normal', truncate=True)
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, pdf='normal', relative=True)
                     0           1
         0 0.00000e+00 0.00000e+00
         1 1.28607e+00 1.10684e+00
         2 1.48457e+00 9.00879e-01
 
-        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, pdf='uniform', truncate=True)
+        >>> sandy.CategoryCov([[1, 0.4],[0.4, 1]]).sampling(3, seed=11, pdf='uniform', relative=True)
```
luca-fiorito-11, line 781: I guess this difference come from the different implementation of the truncation for a uniform distribution.
Then it is ok that the output changes

sandy/core/cov.py
```diff
@@ -745,28 +741,28 @@ def sampling(self, nsmp, seed=None, rows=None, pdf='normal',
         1 1.99199e-01 3.02605e+00
 
         `relative` kwarg usage:
-        >>> sandy.CategoryCov([[1, -.2],[-.2, 3]]).sampling(1000000, seed=11, pdf='normal', tolerance=0, relative=True, truncate=False).data.mean(axis=0)
-        0   1.00014e+00
-        1   9.99350e-01
+        >>> sandy.CategoryCov([[1, -.2],[-.2, 3]]).sampling(1000000, seed=11, pdf='normal', tolerance=0, relative=True).data.mean(axis=0)
```
luca-fiorito-11: To have the same results as before you can do
```
sandy.CategoryCov([[1, -.2],[-.2, 3]]).sampling(1000000, seed=11, pdf='normal', tolerance=0, relative=False).data.mean(axis=0) + 1

```

luca-fiorito-11 (commented):
Few comments on testing and on a variable definition.

luca-fiorito-11:
> Few comments on testing and on a variable definition.

We could already merge but I would like to address these comments before we forget

luca-fiorito-11 (approved):

PR #232, opened 2022-11-19
Sampling ify

EnricaBelfiore:
In this pull request the following methods are changed:

- `sandy.Fy.apply_qmatrix()`: the cut_hl option is inserted to cut all the decay modes of the nuclides with an half life larger than 100 years;

- `sandy.Fy.custom_perturbation()`: adapted to work with a list of ZAP numbers

A Notebook to show the IFYs sampling procedure is added.


sandy/fy.py
```diff
@@ -534,9 +546,12 @@ def apply_qmatrix(self, zam, energy, decay_data, keep_fy_index=False):
             Energy to which calculations are to be applied.
         decay_data : `sandy.DecayData`
             Radioactive nuclide data for several isotopes.
+        cut_hl: `bool`, optional, default is `True`
+            cut all the decay modes of the nuclides with an half life larger
```
luca-fiorito-11: Typo

sandy/fy.py
```diff
@@ -404,11 +404,11 @@ def custom_perturbation(self, zam, mt, e, zap, pert):
         e : `float`
             Energy of the fissioning system to which which perturbations
             are to be applied.
-        zap : `int`
-            ZAP of the product to which perturbations are to be
-            applied.
-        pert : `float`
-            Perturbation coefficients as ratio values.
+        zap : `int` or `list`
```
luca-fiorito-11: I don't like `zap` and `pert` because they are redundant when multiple perturbations are given.
Then, I would rather have `pert` as a list.

Also, the scalar version can be also generalized as a special case of the vector version, with the list containing only one element.

sandy/fy.py
```diff
@@ -422,10 +422,22 @@ def custom_perturbation(self, zam, mt, e, zap, pert):
...
+        >>> assert f1 * 0.5 == comp1
+        >>> comp2 = nfpy_pert.data.query('ZAM==922350 & ZAP==541350 & MT==459 & E==0.0253').squeeze().FY
+        >>> f2 = nfpy.data.query('ZAM==922350 & ZAP==541350 & MT==459 & E==0.0253').squeeze().FY
+        >>> assert f2 * 0.9 == comp2
         """
         df = self.data.copy()
-        mask = (df.ZAM == zam) & (df.ZAP == zap) & (df.MT == mt) & (df.E == e)
-        df.loc[mask, "FY"] = df[mask].squeeze().FY * pert
+        idx = df.query(f"ZAM=={zam} & ZAP=={zap} & MT=={mt} & E=={e}").index
+        fy_pert = df.loc[idx, :].set_index("ZAP").FY * pert
```
luca-fiorito-11, line 436: If pert is a list, then you don't need to match the index anymore for the product and this could be simplified

luca-fiorito-11 (commented):
Some comments on method `custom_perturbation`.
@EnricaBelfiore , I can already merge or wait. Let me know.

sandy/fy.py
EnricaBelfiore: I agree it's redundant, so we can keep `pert` only as a `pd.Series` and remove `zap`. `pert` will be a row of the DataFrame `sandy.Samples.data` in the sampling procedure. If `pert` was a list, the `zap` should be passed in ascending order by the user. If you prefer to have `pert` as a list and two parameters, I will implement it and specify in the docstring that the two should be consistently sorted

luca-fiorito-11: I prefer the list, but it's up to you, considering consistency and maintainability.

luca-fiorito-11: > I agree it's redundant, so we can keep `pert` only as a `pd.Series` and remove `zap`. `pert` will be a row of the DataFrame `sandy.Samples.data` in the sampling procedure. If `pert` was a list, the `zap` should be passed in ascending order by the user. If you prefer to have `pert` as a list and two parameters, I will implement it and specify in the docstring that the two should be consistently sorted

As i see it, `zap` and `pert` should not be in ascending order, right?! They should just be in a consistent order.

luca-fiorito-11: Both ways are fine, but if you go for a series, then I would rather have a `dict`, which is equivalent. 
What do you think?

EnricaBelfiore: My concern is that when filtering `Fy.data` with the list of ZAPs, the index of the resulting sub-DataFrame will be ordered ascendingly independently of the order of the list. This means that the perturbation coefficients should be ordered ascendingly from the one to apply to the nuclide to the lower ZAP. To me, using a data structure that keeps the information together would help. I can agree with the dictionary but the multiplication in line 437 ensures that a perturbation coefficient is multiplied by the fission yield of the corresponding ZAP thanks to the index, so I would have to cast the dictionary to a pd.Series. 

The implementation with the pd.Series is in the next commit

luca-fiorito-11 (approved):

Issue #237, opened 2022-11-28
Issue in `CategoryCov.invert`


luca-fiorito-11:
`CategoryCov.invert` fails for the following example:

```
sandy.get_endf6_file("jeff_33", "xs", 10010).get_errorr(err=1, mt=[102]).get_cov(multigroup=False)
```
with error:
```
Scipy.sparse RuntimeError: Factor is exactly singular
```
A similar issue can be found on stackoverflow at [Scipy.sparse RuntimeError: Factor is exactly singular](https://stackoverflow.com/questions/45683526/python-scipy-sparse-runtimeerror-factor-is-exactly-singular)

PR #238, opened 2022-11-28
CategoryCov invert

luca-fiorito-11:
This PR follows issue #237.

The following modifications were implemented:
 - the algorithm in method `CategoryCov.invert` was replaced with [numpy.linalg.pinv](https://numpy.org/doc/stable/reference/generated/numpy.linalg.pinv.html) for Hermitian matrices .
 - a private method `CategoryCov._double_invert` was implemented for debugging to calculate the inverse of an inverted matrix

### DIFF
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -527,13 +527,13 @@ class CategoryCov():
         """
         Test method get a covariance matrix without negative eigs.
         BEcause of the properties of pinv, this should be equivalent to
-        reconstructing the matrix as `V @ E @ V`, where `V` are the
+        reconstructing the matrix as `V @ E @ V^T`, where `V` are the
         eigenvectors and `E` are truncated eigenvalues.
 
         Returns
         -------
         `sandy.CategoryCov`
-            matrix without problematic eigenvalues
+            inverse of the inverted matrix
 
         Test on real ND covariance.
         >>> c = sandy.get_endf6_file("jeff_33", "xs", 10010).get_errorr(err=1, mt=[102]).get_cov(multigroup=False)
--- a/sandy/core/cov.py
+++ b/sandy/core/cov.py
@@ -474,57 +474,81 @@ class CategoryCov():
             )
         return self.__class__(df)
 
-    def invert(self, rows=None):
+    def invert(self):
         """
         Method for calculating the inverse matrix.
-
-        Parameters
-        ----------
-        tables : `bool`, optional
-            Option to use row calculation for matrix calculations. The
-            default is False.
-        rows : `int`, optional
-            Option to use row calculation for matrix calculations. This option
-            defines the number of lines to be taken into account in each loop.
-            The default is None.
+        If zeros are found on the matrix diagonal, the corresponding rows and columns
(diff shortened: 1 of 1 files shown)


sandy/core/cov.py
```diff
@@ -474,57 +474,80 @@ def get_corr(self):
...
+        -----
+        Many covariance matrices for nuclear data are ill-defined and might
+        have condition numbers that make the matrix inversion process
+        impossible.
+        To make up for this limitation we produce the (Moore-Penrose)
+        pseudo-inverse of a Hermitian matrix, as implemented in `numpy`.
+        Default options are used.
+        This method does not require any pre-processing of the covariance
+        data, e.g. removing zeros from the matrix diagonal or truncating
+        eigenvalues.
```
EnricaBelfiore, line 496: In any case we don't need to truncate the eigenvalues to invert a matrix. `np.linalg.pinv()` allows us to invert singular (even not squared) matrix, while 'np.linalg.inv()' does not

sandy/core/cov.py
```diff
@@ -474,57 +474,80 @@ def get_corr(self):
...
-        M_nonzero_idxs, M_reduce = reduce_size(self.data)
-        cov = sps.csc_matrix(M_reduce.values)
-        rows_ = cov.shape[0] if rows is None else rows
-        data = sparse_tables_inv(cov, rows=rows_)
-        M_inv = restore_size(M_nonzero_idxs, data, len(self.data))
-        M_inv = pd.DataFrame(M_inv.values, index=index, columns=columns)
-        return self.__class__(M_inv)
+        Test method get a covariance matrix without negative eigs.
+        BEcause of the properties of pinv, this should be equivalent to
+        reconstructing the matrix as `V @ E @ V`, where `V` are the
```
EnricaBelfiore: In our case (symmetrical matrix), the matrix is reconstructed as `V @ E @ V^-1` or `V @ E @ V^T` (orthogonal matrix), where `V` is the matrix containing the eigenvectors  and `E` is the diagonal matrix whose diagonal elements are the corresponding eigenvalues. I think that you can still have negative eigenvalues with this decomposition

sandy/core/cov.py
```diff
@@ -474,57 +474,80 @@ def get_corr(self):
...
+        eigenvectors and `E` are truncated eigenvalues.
+
+        Returns
+        -------
+        `sandy.CategoryCov`
+            matrix without problematic eigenvalues
```
EnricaBelfiore: It will return the pinv of the inverted matrix, i.e. the original matrix but it can have the negative eigenvalues which are problematic in our case I think

sandy/core/cov.py
luca-fiorito-11: yes, you still can have the negative eigenvalues

sandy/core/cov.py
luca-fiorito-11: I will change it

luca-fiorito-11:
@EnricaBelfiore , if you accept the changes, can you approve the review?

EnricaBelfiore (approved):

Issue #15, opened 2019-03-13
convert fortran parser to python


luca-fiorito-11:

luca-fiorito-11:
fixed with #267

PR #239, opened 2022-12-07
No more fortran!

makeclean:
Hi, all. I tend to develop on bleeding edge Llinux systems, and thus ride the curve of new compilers and python versions. I've had much pain with the f2py linkage in sandy, and given what it does it was easy to replace with the python package fortran format. So now no more compiling, no f2py and no fortran!

I also updated the unit tests, which would have been failing for anyone due to an early change in return signatures.


luca-fiorito-11:
Thanks for the pull request, much needed!
I'll try and have a look at it in the next few weeks.

luca-fiorito-11:
@EnricaBelfiore , ping

luca-fiorito-11:
If accepted, this should fix issues #15 and #30

SteSeg:
Me and @RemDelaporteMathurin had a look at it. We checked out to @makeclean 's branch and the tests fail because the module `rwf` is not installed. How come the CI passed in the PR?

makeclean:
All the tests fail, or one test fails? I removed the rwf module, which is weird, I wonder how CI passed? My guess is that test isnt covered on the suite - if you note in the test file it only runs `run: pytest --cov=sandy sandy` - where as you (and I) likely ran `pytest` - I wonder how I missed it?!

makeclean:
I'll make the change and re-run :)

RemDelaporteMathurin:
Indeed, I ran the tests with

```
pytest
```

And had a `ModuleNotFound` error because rwf is not installed. This is the [failing line](https://github.com/makeclean/sandy/blob/642b99fbbe5674e94a4b43692e456602468e26af/tests/record_test.py#L11)

However we have a different behaviour with `pytest sandy`.

SteSeg:
Right,  we tried with only `pytest` and everything failed with a `ModuleNotFound` error regarding a rwf module (I guess it is a sort of fortran module). But with `pytest sandy` the only problem with the tests regards the `NJOY` environment variable.

Anyway, I am redoing all the Jupyter Notebooks with this PR and it works fine. The main `develop` branch does not work instead because of the rwf `ModuleNotFound` error.

makeclean:
Yeah, I'll push a fix for this soon - I just forgot to remove that line from the test I think then all will work fine - give me a few days as Im on travel in the US at the moment.

SteSeg:
Sure, no need to hurry. It works fine for me, I just wanted to give the feedback. Have a safe travel!

luca-fiorito-11:
Hi all, as you can see from #249 this discussion triggered a couple of changes in SANDY, one of them being a complete removal of all fortran.
 From what I could see this PR fixed a legacy part of the code (including obsolete tests) that will disappear in the new release.

Depending on the workload I anticipate that the next release won't be available before at least April.
In the meantime, I guess this branch will provide a nice transition for anyone with issues in the fortran compilation.

luca-fiorito-11:
PR #267 was merged into branch [v1.0](https://github.com/luca-fiorito-11/sandy/tree/v1.0). This branch will become the next release.

The PR eliminated any part of the code that was still depending on fortran, practically streamlining the installation of SANDY, especially on windows.

This PR is then obsolete and closed.

PR #252, opened 2023-02-13
Fix sampling unit tests

luca-fiorito-11:
Fixing issue with unit test in `sandy.Samples.test_shapiro` producing different outputs on windows and linux.
The differences are likely to originate from different compilations of the fortran functions called by `scipy.stats.shapiro`.

The test is converted into a notebook.

A different test based on the p-value is implemented instead.

### DIFF
--- a/sandy/core/samples.py
+++ b/sandy/core/samples.py
@@ -132,60 +132,65 @@ class Samples():
 
         Examples
         --------
-        Generate 5000 xs samples normally and log-normally distributed
+        Generate 5000 xs samples normally, log-normally and uniform distributed
         >>> tape = sandy.get_endf6_file("jeff_33", "xs", 10010)
         >>> njoy_kws = dict(err=1, errorr33_kws=dict(mt=102))
         >>> nsmp = 5000
-        >>> smp_norm = tape.get_perturbations(nsmp, njoy_kws=njoy_kws, smp_kws=dict(seed33=1, pdf="normal"))[33]
-        >>> smp_lognorm = tape.get_perturbations(nsmp, njoy_kws=njoy_kws, smp_kws=dict(seed33=1, pdf="lognormal"))[33]
-
-        The Shapiro-Wilks test proves right for the normal samples and the normal distribution.
-        >>> stat_norm = []
-        >>> stat_lognorm = []
-        >>> for nsmp in [10, 50, 100, 500, 1000, 5000]:
-        ...    df = smp_norm.test_shapiro(pdf="normal", size=nsmp)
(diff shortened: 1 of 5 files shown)


luca-fiorito-11:
@GrimFe , if you think this test is ok, then we merge!

sandy/core/samples.py
```diff
@@ -132,60 +132,65 @@ def test_shapiro(self, size=None, pdf="normal"):
```
GrimFe, line 147: I would consider adding a note on the random number generation in Fortran mentioning the behavior we encountered.

sandy/core/samples.py
```diff
@@ -132,60 +132,65 @@ def test_shapiro(self, size=None, pdf="normal"):
...
-        100       9.78776e-01  1.03195e-01        9.40333e-01     6.43714e-41
-        500       9.85038e-01  5.02533e-05        9.40333e-01     6.43714e-41
-        1000      9.89657e-01  1.68679e-06        9.40333e-01     6.43714e-41
-        5000      9.90618e-01  1.03590e-17        9.40333e-01     6.43714e-41
+        >>> smp_norm = tape.get_perturbations(nsmp, njoy_kws=njoy_kws, smp_kws=dict(seed33=seed, pdf="normal"))[33]
+        >>> smp_lognorm = tape.get_perturbations(nsmp, njoy_kws=njoy_kws, smp_kws=dict(seed33=seed, pdf="lognormal"))[33]
+        >>> smp_uniform = tape.get_perturbations(nsmp, njoy_kws=njoy_kws, smp_kws=dict(seed33=seed, pdf="uniform"))[33]
+
+        In this example we defined the following arbitrary convergence criteria:
+            - if the p value is larger than 0.05 we fail to reject the null-hypothesis and we accept the results
```
GrimFe, line 161: Should a reference for the 0.05 criterion be mentioned? To me it seems to be a quite agreed one. I mean, it's a bit more than an "arbitrary convergence criterion", isn't it?

luca-fiorito-11: Not really needed, it is not used by sandy. It is just an example!

sandy/core/samples.py
luca-fiorito-11: The problem is actually that the docstring is incomplete. I will fix it.

Anyway, as I understand there is no issue with the random number generation in Fortran, but rather with the possibly different compilation options.
To me this is not related to the random number generation, hence, I won't comment on it.
Issues with the compilation can be retrieved from browsing this PR and do not need to be necessarily in the docstring.

PR #258, opened 2023-03-21
faster get_cov implementation

luca-fiorito-11:
New implementation
==============
The previous implementation of `Errorr.get_cov` was:

- reading covariance blocks from ERRORR output file
- stacking covariance data into a 3-column (index, column, value) dataframe
- concatenate all stacked covariance blocks
- `pivot_table` to  reconstruct a global covariance matrix taking into account of all blocks.

This implementation suffered from slow running times.

A new implementation is proposed with:

- a preallocation of the global covariance matrix based on prior size knowledge
- reading covariance blocks from ERRORR output file
- placing covariance blocks in the correct position in the global matrix

Other info
=======

- The keyword option `multigroup` of `get_cov` is removed. Multigroup processing is now considered as default.
- A notebook is added to combine covariance matrices for different materials (pu239 and pu240) and data types (xs and nubar).


luca-fiorito-11:
@GrimFe , have a look!

luca-fiorito-11:
issue with PR, will resubmit

PR #261, opened 2023-03-27
Lbarone2

Luiigi99:
Methods for mean and standard deviation convergence modified in a way that the results are coherent with samples class


Luiigi99:
Wrong

PR #265, opened 2023-05-03
Fix meta stables

fdcroce:
Fix typo error in method njoy.process_neutrons() and implementation of a new test in endf6.get_ace() to check it works correctly.
Edit worker endf6.endf_pendf_worker to take the correct ZA to emply in the name of ace file. generated The previous version was not able to name correctly in case of metastable nuclides.


sandy/core/endf6.py
```diff
@@ -2219,12 +2224,15 @@ def endf6_perturb_worker(e6, pendf, n,
     intro = endf6_pert.read_section(mat, 1, 451)
     za = int(intro["ZA"])
     meta = int(intro["LISO"])
-    zam = sandy.zam.za2zam(za, meta=meta)
+    zam = sandy.zam.za2zam(za, meta=meta,method="")
```
luca-fiorito-11: This change is not needed

sandy/core/endf6.py
```diff
@@ -2219,12 +2224,15 @@ def endf6_perturb_worker(e6, pendf, n,
...
+    zaid = ace_kws.get("zaid", "nndc")
+    if zaid == "nndc":
+        za = sandy.zam.zam2za(zam, method=zaid)[0]
     params = dict(
         MAT=mat,
         ZAM=zam,
-        ZA=za,
         META=meta,
+        ZA = za,
```
luca-fiorito-11: No spaces

luca-fiorito-11 (changes requested):
Two minor comments

luca-fiorito-11:
Can you also comment out the failing test? We'll address it later

sandy/core/endf6.py
fdcroce: It is. The za retrieved the first part of endf6 file doesn't follow _"nndc"_, then if method za2zam is _"nndc"_ as default  meta information is discarded. This brings to generation of wrong zam.
Maybe better to change notation in _method=False_ consistently with examples in zam module.

luca-fiorito-11: ok, but add a space between "," and "method"

luca-fiorito-11 (approved):

luca-fiorito-11 (approved):

PR #269, opened 2023-05-25
Energy distr

Luiigi99:
Modification of _errorr_input to work also with ifissp ( incident energy group number of requested covariance matrix) and efmean (incident energy at which covariance matrix is evaluated).


luca-fiorito-11:
outdated

PR #270, opened 2023-05-25
Lbarone4

Luiigi99:
Modifications to work correctly with nubar. In this way samples of perturbed nubar and/or cross sections can be obtained (also simultaneously if wanted)


luca-fiorito-11:
inactive

PR #271, opened 2023-05-25
Fix covariance matrices symmetry

fdcroce:
Update of the method to check symmetry of covariance matrix, setting the _max deviation_ among the covariance matrix and its transposed. 
Moreover a minor edit was carried out in the method _get_cov()_. When adding the diagonal sections to the full covariance matrix is not necessary to add also the transposed.


sandy/core/cov.py
```diff
@@ -311,9 +311,8 @@ def data(self, data):
         if not (np.diag(data) >= 0).all():
             raise TypeError("Covariance matrix must have positive variance")
 
-        sym_limit = 10
-        # Round to avoid numerical fluctuations
-        if not (data.values.round(sym_limit) == data.values.T.round(sym_limit)).all():
+        treshold = 1e-7
```
luca-fiorito-11, line 314: Typo

sandy/errorr.py
```diff
@@ -183,8 +197,9 @@ def get_cov(self):
                     jvals = ix.query("MAT==@mat & MT==@mt1").squeeze()
                     jmin, jmax = jvals.IMIN, jvals.IMAX
                     c[imin: imax, jmin: jmax] = cov
-                    c[jmin: jmax, imin: imax] = cov.T
-            
+                    if mt1 != mt:
```
luca-fiorito-11, line 200: Is this needed?

sandy/errorr.py
```diff
@@ -315,7 +330,7 @@ def read_mf33(tape, mat, mt):
             GROW = L.N2
             GCOL = L.L2
             M[GROW-1, GCOL-1:GCOL+NGCOL-1] = L.B
-            if GCOL+NGCOL >= NG and GROW >= NG:
+            if GROW >= NG:
                 break
```
luca-fiorito-11, line 334: I thought this was already fixed

luca-fiorito-11 (approved):
Can you just fix the typo

luca-fiorito-11:
@fdcroce , I see there are conflicts in errorr. Can you fix them?

sandy/errorr.py
fdcroce: If the section of covariance matrix is already in diagonal position line 186 is useless, but it doesn't create issues so I can also leave it as it was.

sandy/errorr.py
fdcroce: Yes it was in my previous PR. I think this and the conflicts in errorr come from the fact I didn't branch directly from v1.0.
Maybe I'm creating the same branch but from v1.0 so we solve the issues. Can you reject this one. I'm opening a new PR following the standard procedure.

luca-fiorito-11: Will close it then

PR #280, opened 2023-08-10
Faster rdd

luca-fiorito-11:
adding decay data files locally for faster parsing.

this should also fix issues with notebooks.


luca-fiorito-11:
need to remove pytables, because new issue with pandas: [pandas/issues/54449](https://github.com/pandas-dev/pandas/issues/54449).

PR #283, opened 2023-09-05
Energy distribution implementation

Luiigi99:
The branch contains the methods needed to read, manage and perturb also the energy distribution data. A new class Edistr has been created according to the agreed scheme.

P.S. some methods(e.g. get_table and add_energy_points) of the new class Edistr have not been modified because unnecessary for the perturbation process.


luca-fiorito-11:
outdated

Issue #287, opened 2023-09-21
Read perturbations coefficients from file


luca-fiorito-11:
Reading perturbation coefficients that were previously generated ensures reproducibility and could save disk space (no need to store all samples all the times). Also, only a fraction of samples could be generated from the perturbation coefficients, leaving a second fraction for later.

luca-fiorito-11:
Needs to install openpyxl to read/write excel files

luca-fiorito-11:
close in #288

Issue #286, opened 2023-09-21
Print covariance eigenvalues and eigenvectors in sampling procedure


luca-fiorito-11:
The sampling procedure shall give the user the possibility to access the sampled coefficient, but also the eigenvalues and eigenvectors of the original covariance matrix.

luca-fiorito-11:
closed in #288

PR #288, opened 2023-09-21
Read/write perturbation coefficients to file

luca-fiorito-11:


luca-fiorito-11:
Closed #287 and #286

luca-fiorito-11:
Added openpyxl to config

PR #304, opened 2024-03-23
Add fstart argument to write_line

janmalec:
It is needed to control how the pendf files are put together. Just like fend, which is already implemented.

### DIFF
--- a/sandy/core/endf6.py
+++ b/sandy/core/endf6.py
@@ -1206,7 +1206,7 @@ class _FormattedFile():
 
         How to use keyword `tpid`.
         >>> first = endf6.write_string(tpid=False).splitlines()[0]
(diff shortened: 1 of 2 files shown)


luca-fiorito-11:
It makes sense. 
However, any addition to the code requires its own test in the docstring. Can you add it?
Also, FEND is ENDF6 nomenclature. Can you check if fstart is part of the default nomenclature?
If no indication is given in the endf6 manual, then we can accept fstart

janmalec:
I changed FSTART to TPID, which is ENDF6 nomenclature and added docstring tests.

luca-fiorito-11 (approved):

luca-fiorito-11:
You submitted two items to the same PR. Can you submit two PRs instead?

luca-fiorito-11:
@janmalec , the test failed. Can you check it?

PR #306, opened 2024-05-02
Fix write_float edge case, which crashes NJOY

AnderGray:
There is a strange edge case in `write_float` where

```python
sandy.records.write_float(0.99999995) 
' 9.999999-1'
```
but
```python
sandy.records.write_float(0.99999996) 
' 1.000000e+00'
```

Likely because the string interpolation `f"{x:13.6e}"` is rounding up to ` 1.000000e+00`, and it isn't caught as a case in the function. 

This was giving an annoying error from NJOY `Fortran runtime error: Bad value during floating point read`, where these strings end up in generated pendf files, and are unable to be read by NJOY.

### DIFF
--- a/sandy/core/records.py
+++ b/sandy/core/records.py
@@ -315,7 +315,7 @@ def write_float(x):
         y = f"{x:11.0f}"
     elif x == 0:
         y = f"{x:11.8f}"
-    elif abs(x) < 1E0 and abs(x) >= 0.99999996:
+    elif abs(x) < 1E0 and abs(x) > 0.99999995:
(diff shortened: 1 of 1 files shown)


luca-fiorito-11:
This was fixed (also in principle tested) by @janmalec in #308 as an answer to issue #307.
The changes were pushed to branch v1.1 because we don't touch develop directly.

I guess a release will be issued for v1.1 within the next few months, and only then it will be merged back into develop.

Btw, thanks for spotting this problem.
I invite you to check if v1.1 works fine for you, or else give feedback!

PR #315, opened 2024-05-30
Adding arguments to CLI to manage terminal outputs

nplinden:
The purpose of the PR is to allow the user of `python -m sandy.sampling` to suppress NJOY output using the `--supressnjoy` argument, and to set the logging level through the `--loglevel` arguments.


luca-fiorito-11:
I delete the short version '-s' of '--suppressnjoy'. I think it can be confusing with '-S'

luca-fiorito-11:
I didn't implement unit tests because i don't see how to do it

PR #379, opened 2025-06-11
cherrypick_workflow

luca-fiorito-11:


luca-fiorito-11:
have to merge even if CI failed

PR #402, opened 2025-07-03
Allow empty MODn value as zeros in MF1 MT451

### LINKED ISSUE #401: Processing JEFF-4 U235 tape
Hello,
I was trying to process the new U235 tape from JEFF-4 with sandy and I get the following error `ValueError: invalid literal for int() with base 10: '           '` when reading the table of content part of MF1 MT451.

It seems that the reason is that at line 734 of the tape the value of MODn is empty rather than being zero:

```
                               12         79          5          19228 1451  720
                               12         80          3          19228 1451  721
                               12         81          3          19228 1451  722
                               12         82          4           9228 1451  294 <- This won't work
                               12         83          4           9228 1451  295 <- This won't work
                               12         84          4           9228 1451  296 <- This won't work
                               12         85          4           9228 1451  297 <- This won't work
                               12         86          4           9228 1451  298 <- This won't work
                               12         87          4           9228 1451  299 <- This won't work
                               12         88          4           9228 1451  300 <- This won't work
                               12         89          5           9228 1451  301 <- This won't work
                               12         90          5           9228 1451  302 <- This won't work
                               14         18          1          19228 1451  732
                               14         51          1          19228 1451  733

```

I guess the tape is malformed, but this wasn't caught by the NEA pipeline for some reason? Maybe the ENDF6 standard accepts no value as zero is this case?

The easiest fix would be to modify sandy's `read_mf1` function to cast the empty lines to 0. The tape is read normally by IAEA's endf-parserpy and the empty values are indeed replaced with zeros.

nplinden:
This is a simple fix for this #401 issue. Not sure how to create a test for this as the bug is seen on JEFF-4 which are not yet available with get_endf6_file()

### DIFF
--- a/sandy/sections/mf1.py
+++ b/sandy/sections/mf1.py
@@ -480,6 +480,7 @@ def _read_intro(tape, mat):
     sections = []
     for j in range(NXC):
         T, i = sandy.read_text(df, i)
+        # if MOD is left empty, add a 0 
         s = tuple(map(lambda x: 0 if x.strip() == "" else int(x), 
                       re.findall(".{11}", T[0])[2:])
                   )
--- a/sandy/sections/mf1.py
+++ b/sandy/sections/mf1.py
@@ -480,7 +480,9 @@ def _read_intro(tape, mat):
     sections = []
     for j in range(NXC):
         T, i = sandy.read_text(df, i)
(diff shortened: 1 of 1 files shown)


luca-fiorito-11:
Thanks!

